PR #6, opened 2026-04-12
Implement aggregate functions with GROUP BY support

ragnorc:
## Summary
This PR adds support for aggregate functions (COUNT, SUM, AVG, MIN, MAX) with GROUP BY semantics to the query execution engine. It also refactors the graph traversal (Expand) operation to maintain row alignment across correlated bindings, which is essential for correct aggregate computation.

### Aggregate Function Support
- Added `aggregate_return()` function to handle queries with aggregate expressions
- Implements grouping logic that classifies projections into group-by keys (non-aggregate expressions) and aggregate expressions
- Supports all five aggregate functions with proper type handling:
  - `COUNT`: counts non-null values per group
  - `SUM`: sums numeric values (Int32/64, UInt32/64, Float32/64)
  - `AVG`: computes average as Float64
  - `MIN/MAX`: finds extrema with support for numeric and string types
- Handles edge case of empty input (returns single row with COUNT=0, others=NULL)
- Added type checking rule T9 to ensure non-aggregate expressions in aggregate queries are valid group-by keys (PropAccess or Variable only)

### Row Alignment for Expand Operations
- Refactored `execute_expand()` to mutate bindings in-place and maintain row alignment
- Changed return type from `Result<RecordBatch>` to `Result<()>` with direct binding updates
- Implements join-aligned semantics: for each (source_row, dest_node) pair, all existing bindings are repeated so row `i` across all bindings represents one matched tuple
- Uses index arrays (`src_indices`, `dst_indices`) to perform aligned `take` operations on all correlated bindings
- Sources with zero destinations are dropped (inner-join semantics)

### Filter and AntiJoin Alignment
- Updated `apply_filter()` to filter all correlated bindings (same row count) to maintain alignment
- Enhanced `execute_anti_join()` to filter all correlated bindings when removing rows
- Added fast-path optimization in `execute_anti_join()` that reconstructs the filter mask for other bindings

### Compiler Changes
- Added `has_aggregate` flag to `QueryIR` struct to track if query contains aggregates
- Updated query lowering to propagate aggregate information
- Added `AggFunc` import to execution module

### Testing
- Added comprehensive test suite (`aggregation.rs`) covering:
  - COUNT with GROUP BY (friend counts per person)
  - Global COUNT (no group key)
  - SUM/AVG/MIN/MAX with GROUP BY (age statistics per company)
  - Aggregate with ORDER BY and LIMIT (top connected person)
- Added test queries to fixture file

## Implementation Details
- Group keys are built as delimited strings (using `\x1F` unit separator) to handle multi-column grouping
- Null values in group keys are represented as `\x00NULL` to distinguish from empty strings
- Aggregate computation iterates through grouped row indices, handling nulls appropriately per function
- Result columns are sorted back to original projection order after computation
- Variable projection support added to `evaluate_projection()` for accessing node IDs in aggregates

https://claude.ai/code/session_019o5NRyYomgETFyd7hpiLey

---

> [!NOTE]
> **Medium Risk**
> Adds new aggregation execution path and refactors the query pipeline to operate on a single wide `RecordBatch` with prefixed columns, changing traversal/antijoin/filter semantics and ordering resolution; this touches core query execution and could impact result correctness/performance.
> 
> **Overview**
> Enables aggregate queries (`count`, `sum`, `avg`, `min`, `max`) with implicit GROUP BY semantics by adding an aggregate projection path that groups by non-aggregate return expressions and computes per-group results (including an empty-input global-aggregate case).
> 
> Refactors execution from per-variable `bindings` to a single *wide* batch with `var.property` column names: node scans are cross-joined, `expand` now maintains row alignment by duplicating source rows per destination and horizontally concatenating prefixed destination columns, and filters/antijoins operate on the wide batch. Ordering is updated to resolve property accesses against either the wide batch or the aggregate result batch.
> 
> Tightens compiler/typechecking for aggregates: enforces that non-aggregate projections in aggregate queries are valid group keys (only prop accesses or variables) and allows `min`/`max` over strings; result schema inference updates `sum`/`avg` to `Float64` and `min`/`max` to the argument type. Adds an aggregation-focused test suite and new fixture queries covering grouped/global aggregates plus ordering/limit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 351610d18c105862fc8702a8164e381c30dd8a4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/omnigraph-compiler/src/query/typecheck.rs
+++ b/crates/omnigraph-compiler/src/query/typecheck.rs
@@ -189,6 +189,29 @@ fn typecheck_read_query(catalog: &Catalog, query: &QueryDecl) -> Result<TypeCont
         ));
     }
 
+    // T9: If any return expression is an aggregate, non-aggregate expressions
+    // must be valid group-by keys (PropAccess or Variable).
+    let has_agg = query
+        .return_clause
+        .iter()
+        .any(|p| matches!(p.expr, Expr::Aggregate { .. }));
+    if has_agg {
+        for proj in &query.return_clause {
+            if !matches!(proj.expr, Expr::Aggregate { .. }) {
+                match &proj.expr {
+                    Expr::PropAccess { .. } | Expr::Variable(_) => {}
+                    _ => {
+                        return Err(NanoError::Type(
+                            "T9: non-aggregate expressions in an aggregate query must be \
+                             property accesses or variables"
+                                .to_string(),
+                        ));
+                    }
+                }
+            }
+        }
+    }
+
     Ok(ctx)
 }
 
@@ -1298,14 +1321,25 @@ fn resolve_expr_type(
         Expr::Aggregate { func, arg } => {
             let arg_type = resolve_expr_type(catalog, arg, ctx, params)?;
 
-            // T8: sum/avg/min/max require numeric
+            // T8: sum/avg require numeric; min/max require numeric or string
             match func {
-                AggFunc::Sum | AggFunc::Avg | AggFunc::Min | AggFunc::Max => {
+                AggFunc::Sum | AggFunc::Avg => {
                     if let ResolvedType::Scalar(s) = &arg_type
                         && (s.list || !s.scalar.is_numeric())
                     {
                         return Err(NanoError::Type(format!(
                             "T8: {} requires numeric type, got {}",
+                            func,
+                            s.display_name()
+                        )));
+                    }
+                }
+                AggFunc::Min | AggFunc::Max => {
+                    if let ResolvedType::Scalar(s) = &arg_type
+                        && (s.list || (!s.scalar.is_numeric() && s.scalar != ScalarType::String))
+                    {
+                        return Err(NanoError::Type(format!(
+                            "T8: {} requires numeric or string type, got {}",
                             func,
                             s.display_name()
                         )));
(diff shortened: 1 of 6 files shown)


aaltshuler (commented):
Three issues from the aggregate patch need attention:

1. Variable projections no longer match their declared result type. The new `IRExpr::Variable` projection path returns the `<binding>.id` UTF-8 column from the wide batch, but the compiler still resolves `Expr::Variable` to `ResolvedType::Node(...)` and advertises a full struct via `resolved_type_to_field_shape`. Queries like `return { $p }` will therefore typecheck as a node-shaped result and execute as a string ID column instead.

2. `sum(...)` output type diverges from the compiler contract. `compute_sum` now always materializes a `Float64Array`, but projection typing still uses the aggregate argument's type for every aggregate other than `count` and `avg`. A query such as `return { sum($p.age) }` will therefore be typed as `Int32`/`Int64` at compile time and returned as `Float64` at runtime.

3. `min`/`max` on strings are still rejected by typechecking. The executor adds `Utf8` handling in `compute_min_max`, but T8 in `resolve_expr_type` still groups `min` and `max` with `sum` and `avg` under a numeric-only check. As written, `min($n.name)` and `max($n.name)` still fail in the compiler even though the runtime support was added.

PR #22, opened 2026-04-16
fix: build Lance indexes locally when repo is S3-backed (MR-640)

ragnorc:
## Summary

Fixes [MR-640](https://linear.app/modernrelay/issue/MR-640/rustfs-http-body-error-on-lance-ranged-reads-above-8k-rows): Lance index creation fails on RustFS for larger datasets because it reads column data back from S3 via ranged GETs, and RustFS breaks HTTP response body streaming.

**Design flaw**: Lance builds indexes by reading back data it just wrote to the same remote store. The data round-trips through the network for no reason.

**Fix**: Stage all large S3 reads through local filesystem. Download dataset files via individual object GETs (small files), scan/index locally, upload finished files back to S3 as simple PUTs.

## Changes

**`table_ops.rs`**:
- `prepare_updates_for_commit`: skip index building for S3 repos (deferred to `ensure_indices` after data commit)
- `ensure_indices_for_branch`: use `build_indices_via_local_staging` for S3 (download → local index → upload)
- New: `build_indices_via_local_staging`, `index_staging_tempdir`, `is_s3_repo`

**`omnigraph.rs`**:
- New: `s3_safe_batch_for_table_rewrite` — downloads dataset before scanning for S3 repos
- `promote_run_snapshot_to_target` + `reify_internal_run_refs`: use S3-safe scan

**`table_store.rs`**:
- New: `copy_remote_dataset_to_local` / `copy_local_dataset_to_remote`

**`loader/mod.rs`**:
- `load()`: call `ensure_indices_on` after publish for S3 repos

## Test plan

- [x] `s3_large_load_builds_indices_without_error` — 14K rows, deterministic FAIL before fix, PASS after
- [x] Full S3 test suite (4/4 pass)
- [x] Full workspace tests pass (all suites, 0 failures)
- [x] Diff is clean — only MR-640 changes, no unrelated code

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the write/publish/indexing workflow for S3 repos and adds new dataset file copy logic that could impact correctness or performance of remote table state and indices.
> 
> **Overview**
> Fixes S3/RustFS failures during Lance scans and index creation by **staging remote datasets locally**: download dataset files to a temp dir, scan/build indexes against the local copy, then upload the resulting dataset (including index files) back to S3.
> 
> This switches `publish_run` table rewrites to use `s3_safe_batch_for_table_rewrite`, updates `ensure_indices` to use `build_indices_via_local_staging` on S3, and **defers index creation during commit** for S3 (building them post-publish in `load()` and transactional `mutate()` with non-fatal warnings on failure). Adds `TableStore::copy_remote_dataset_to_local` / `copy_local_dataset_to_remote`, plus new S3 regression tests (including a large 14K-row fixture) to validate indexes are created successfully.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a314f6a9d1b2c182b9245337f33694e68c40634e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/omnigraph/src/db/omnigraph.rs
+++ b/crates/omnigraph/src/db/omnigraph.rs
@@ -910,7 +910,7 @@ impl Omnigraph {
             };
 
             let source_ds = run_snapshot.open(&table_key).await?;
-            let batch = self.batch_for_table_rewrite(&source_ds, &table_key).await?;
+            let batch = self.s3_safe_batch_for_table_rewrite(&source_ds, &table_key).await?;
 
             let (mut target_ds, full_path, table_branch) = self
                 .open_for_mutation_on_branch(target_branch.as_deref(), &table_key)
@@ -959,7 +959,7 @@ impl Omnigraph {
 
             let source_ds = target_snapshot.open(&entry.table_key).await?;
             let batch = self
-                .batch_for_table_rewrite(&source_ds, &entry.table_key)
+                .s3_safe_batch_for_table_rewrite(&source_ds, &entry.table_key)
                 .await?;
 
             let (mut target_ds, full_path, table_branch) = self
@@ -1120,6 +1120,45 @@ impl Omnigraph {
         table_key: &str,
     ) -> Result<RecordBatch> {
         schema_apply::batch_for_table_rewrite(self, source_ds, table_key).await
+    }
+
+    /// Like `batch_for_table_rewrite`, but when the repo is S3-backed,
+    /// downloads the dataset files to local before scanning. This avoids
+    /// Lance's concurrent ranged GETs over HTTP which fail on RustFS.
+    async fn s3_safe_batch_for_table_rewrite(
+        &self,
+        source_ds: &Dataset,
+        table_key: &str,
+    ) -> Result<RecordBatch> {
+        use crate::storage::{StorageKind, storage_kind_for_uri};
+        use crate::table_store::TableStore;
+
+        if storage_kind_for_uri(&self.root_uri) != StorageKind::S3 {
+            return self.batch_for_table_rewrite(source_ds, table_key).await;
+        }
+
+        let row_count = self
+            .table_store()
+            .count_rows(source_ds, None)
+            .await
+            .unwrap_or(0);
+        if row_count == 0 {
+            return self.batch_for_table_rewrite(source_ds, table_key).await;
+        }
+
+        let staging = tempfile::Builder::new()
+            .prefix(&format!("omnigraph-scan-{}-", table_key.replace(':', "-")))
+            .tempdir()
+            .map_err(OmniError::from)?;
+        let local_uri = staging.path().join("dataset").to_string_lossy().to_string();
+
+        TableStore::copy_remote_dataset_to_local(source_ds.uri(), &local_uri).await?;
+
+        let local_ds = Dataset::open(&local_uri)
+            .await
+            .map_err(|e| OmniError::Lance(format!("open local staged dataset: {}", e)))?;
+        self.batch_for_table_rewrite(&local_ds, table_key).await
+        // TempDir cleaned on drop.
     }
 }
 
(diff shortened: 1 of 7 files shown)


aaltshuler:
Closing — this is a ducktape fix layered over a real upstream bug we should actually fix at its source.

MR-640's investigation confirms:

- MinIO handles the identical workload without error
- Local FS handles it without error
- Only RustFS fails at ~8K rows, due to broken HTTP/1.1 response body streaming on ranged GETs

Since RustFS is our on-prem target, the right fix is upstream — not re-architecting our data path around it. This PR adds 771 LOC across 7 files, routes every S3 dataset through a local disk round-trip for every index build, introduces a "best-effort silent failure" path on post-publish, and still leaves the schema-apply path broken.

## Proposed next steps

1. **Further investigation into the failure mode.** Wire-level capture of the failing request/response (chunked-encoding trailer? premature RST? malformed Content-Length?). MR-640's table documents the symptom; we need protocol-level detail to file an actionable bug.
2. **File the bug upstream against RustFS** with the wire-level evidence. They can't fix what they don't know is broken.
3. **Temporarily swap the dev bootstrap to MinIO** while the upstream fix is in flight. Contributors shouldn't be blocked by a RustFS bug they didn't cause.
4. **Extract the "indexes built twice" finding** from MR-640 into its own small PR — remove the redundant run-phase `build_indices_on_dataset` calls at `omnigraph.rs:586/632`. That's real simplification, independent of the RustFS bug.

aaltshuler:
Follow-up repro attempt — bug does **not** reproduce on native Mac RustFS.

## Environment

| | |
|---|---|
| RustFS | `1.0.0-alpha.90`, native Mac binary, build 2026-03-24 |
| Host | Darwin 25.1.0, Apple Silicon |
| Client | Lance 4.0.0, object_store 0.12.5 (unchanged from MR-640) |
| Launch | `rustfs --address 127.0.0.1:9000 --console-address 127.0.0.1:9099 rustfs-data` |

## Test matrix

| Test | Rows | Data file size | Result |
|---|---|---|---|
| `omnigraph init` | — | — | OK |
| `omnigraph load --mode overwrite` | 10,000 | 4.8 MB JSONL | OK |
| `omnigraph load --mode overwrite` | 30,000 | 15 MB JSONL | OK (3× MR-640 threshold) |
| BTree index on `id` | — | 99,810 B `page_data.lance` | built |
| FTS indexes on `name`, `kind`, `source`, `timestamp` | — | — | all built |

Schema mirrors MR-640's generator: single `Artifact` node type, wide indexed string fields, 50–250 char padded `content`, indexed enums `kind`/`source`, indexed `timestamp`.

Zero errors. No `request or response body error`, no Lance scheduler panic, no HTTP failures.

## Interpretation

MR-640 tested `rustfs/rustfs:latest` Docker image on 2026-04-15. We tested native Mac `alpha.90` on 2026-04-17. The failure mode is therefore one or more of:

1. **Docker-specific** — container HTTP stack, network shim, cgroup limits.
2. **Version-specific** — fixed between `latest` on 2026-04-15 and `alpha.90`.
3. **Platform-specific** — Linux (Docker) vs Darwin (native) handle HTTP body streaming differently.

The bug is not a universal RustFS property, and this PR's workaround was never retested against the actual on-prem target binary.

## Bonus finding — "indexes built twice" confirmed

The S3 layout after one `omnigraph load` shows indexes in both places:

```
nodes/fa22f663baf34923/_indices/                    ← 5 UUIDs on main
nodes/fa22f663baf34923/tree/__run__.../_indices/    ← 5 UUIDs on run branch
```

Run-branch indexes are rebuilt on main during publish and discarded. That's a real, independent simplification — worth extracting as a focused PR regardless of the RustFS question.

## Recommendations

1. **Re-verify against the on-prem target binary.** If ModernRelay ships native (not Docker), the bug may not be real in production.
2. **If Docker is a supported deployment**, file a narrow bug against the RustFS Docker image/version with a wire-level trace (tcpdump / mitmproxy).
3. **Extract the "indexes built twice" cleanup** independently. Valuable regardless of the RustFS issue.

PR #51, opened 2026-04-25
Prefix commit and run IDs with cmt_ / run_

ragnorc:
## Summary

Surface ULIDs through the API with type-tag prefixes:

| Before | After |
|---|---|
| `01KQ2JK3PC35ZZW3867A5H7Q15` | `cmt_01KQ2JK3PC35ZZW3867A5H7Q15` (commit) |
| `01KQ2KB5KFPNB7KMFM7816S6J2` | `run_01KQ2KB5KFPNB7KMFM7816S6J2` (run) |

Logs, dashboards, and bug reports now disambiguate the two ID families at a glance. Snapshot IDs also pick up the `cmt_` prefix since they refer to commit IDs.

This is **Part A2** of the Stripe-style TypeScript SDK rollout (omnigraph-ts#2).

## Where prefixing happens

Storage stays on bare ULIDs — no migration of existing data. Prefixing happens at the API serialization boundary in `commit_output()`, `run_output()`, and `read_target_output()`. Inputs accept either form (`unprefix_commit_id()` and `unprefix_run_id()` strip the prefix before storage lookups), so callers holding old-format IDs from logs continue to work.

## Test plan

- [x] `cargo test -p omnigraph-server --lib api::tests` — 8 unit tests covering: prefix adds, prefix is idempotent, unprefix strips, unprefix passes through bare ULID
- [x] `cargo test -p omnigraph-server` — 41 server tests pass (`policy_uses_resolved_branch_for_snapshot_reads` updated to accept prefixed snapshot id)
- [x] `cargo test -p omnigraph-cli` — 85 CLI tests pass (`run_publish_*` and `run_abort_*` updated to assert prefixed return values)
- [x] `cargo test -p omnigraph-server --test openapi` — 65 OpenAPI tests pass; no spec drift (prefixing is a value-level concern, not a schema change)

## Companion CLI changes

The CLI's local-path commit/run commands now strip prefixes before hitting `db.get_commit` / `RunId::new` so users can paste the displayed `cmt_…` or `run_…` IDs back into `omnigraph commit show` and `omnigraph run show/publish/abort`.

---

> [!NOTE]
> **Medium Risk**
> Changes API/CLI-visible identifier formats for commits, runs, and snapshot targets, which can break downstream consumers that assume bare ULIDs despite input-side compatibility shims.
> 
> **Overview**
> API responses now **prefix commit, run, and snapshot IDs** with type tags (`cmt_…` for commits/snapshots, `run_…` for runs) via new helpers in `omnigraph-server` (`prefix_*`/`unprefix_*`) applied in `commit_output`, `run_output`, and `read_target_output`.
> 
> Server routes and the CLI now **accept both prefixed and bare IDs** by stripping prefixes before DB lookups (e.g., run/commit show, run publish/abort, and snapshot read targets), and tests are updated to assert the new prefixed values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f0050092b9e6c23800500fb4fb56b051a9ed096. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Prefix API-visible IDs to disambiguate types: commit IDs now use `cmt_…`, run IDs use `run_…`, and snapshot IDs use `cmt_…`. Storage stays on bare ULIDs, and inputs accept both prefixed and bare forms.

- **New Features**
  - `omnigraph-server` responses prefix commit (`cmt_…`) and run (`run_…`) IDs; parent/merged parent commit IDs, base/published snapshot IDs, and read-target snapshot IDs are also prefixed.
  - Prefixing happens at serialization; server routes and `omnigraph-cli` strip prefixes on input for backward compatibility.

- **Bug Fixes**
  - `omnigraph-cli` `read --snapshot` now strips `cmt_` so pasted IDs work locally; server test tightened to require prefixed snapshot IDs in responses.

<sup>Written for commit 5f0050092b9e6c23800500fb4fb56b051a9ed096. Summary will update on new commits.</sup>

### DIFF
--- a/crates/omnigraph-cli/src/main.rs
+++ b/crates/omnigraph-cli/src/main.rs
@@ -20,7 +20,7 @@ use omnigraph_server::api::{
     CommitOutput, ErrorOutput, ExportRequest, IngestOutput, IngestRequest, ReadOutput, ReadRequest,
     RunListOutput, RunOutput, SchemaApplyOutput, SchemaApplyRequest, SchemaOutput, SnapshotOutput,
     SnapshotTableOutput, commit_output, ingest_output, read_output, run_output,
-    schema_apply_output, snapshot_payload,
+    schema_apply_output, snapshot_payload, unprefix_commit_id, unprefix_run_id,
 };
 use omnigraph_server::{
     AliasCommand, OmnigraphConfig, PolicyAction, PolicyDecision, PolicyEngine, PolicyRequest,
@@ -2006,7 +2006,7 @@ async fn main() -> Result<()> {
                     .await?
                 } else {
                     let db = Omnigraph::open(&uri).await?;
-                    commit_output(&db.get_commit(&commit_id).await?)
+                    commit_output(&db.get_commit(unprefix_commit_id(&commit_id)).await?)
                 };
                 if json {
                     print_json(&commit)?;
@@ -2247,7 +2247,10 @@ async fn main() -> Result<()> {
                     .await?
                 } else {
                     let db = Omnigraph::open(&uri).await?;
-                    run_output(&db.get_run(&RunId::new(run_id)).await?)
+                    run_output(
+                        &db.get_run(&RunId::new(unprefix_run_id(&run_id).to_owned()))
+                            .await?,
+                    )
                 };
                 if json {
                     print_json(&run)?;
@@ -2277,8 +2280,9 @@ async fn main() -> Result<()> {
                     .await?
                 } else {
                     let mut db = Omnigraph::open(&uri).await?;
-                    db.publish_run(&RunId::new(run_id.clone())).await?;
-                    run_output(&db.get_run(&RunId::new(run_id)).await?)
+                    let bare = unprefix_run_id(&run_id).to_owned();
+                    db.publish_run(&RunId::new(bare.clone())).await?;
+                    run_output(&db.get_run(&RunId::new(bare)).await?)
                 };
                 if json {
                     print_json(&run)?;
@@ -2308,7 +2312,10 @@ async fn main() -> Result<()> {
                     .await?
                 } else {
                     let mut db = Omnigraph::open(&uri).await?;
-                    run_output(&db.abort_run(&RunId::new(run_id)).await?)
+                    run_output(
+                        &db.abort_run(&RunId::new(unprefix_run_id(&run_id).to_owned()))
+                            .await?,
+                    )
                 };
                 if json {
                     print_json(&run)?;
--- a/crates/omnigraph-cli/src/main.rs
+++ b/crates/omnigraph-cli/src/main.rs
@@ -1444,7 +1444,7 @@ fn inferred_config_path(uri: &str) -> Result<PathBuf> {
 
 fn read_target_from_cli(branch: Option<String>, snapshot: Option<String>) -> ReadTarget {
     if let Some(snapshot) = snapshot {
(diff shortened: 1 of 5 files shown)


aaltshuler:
Pushing back on the approach. I get the Stripe-style appeal, but this design locks us into three permanent costs:

**1. Two representations forever.** `01K…` and `cmt_01K…` both refer to the same commit because inputs accept both. Any consumer that stores an ID and later compares it against one reconstructed elsewhere hits silent equality bugs. Caches, sets, dedup — all get two keys for the same thing. The only way out is being strict on input, which kills the "old log lines still work" ergonomic that justified the prefix in the first place.

**2. Boundary discipline forever.** Every new HTTP handler, every new output struct has to remember to call `unprefix_*` / `prefix_*`. Nothing in the type system enforces it — both sides are `String`. The first time someone forgets, we ship an inconsistent response or silently accept the wrong-typed ID and nobody notices until a user reports something weird.

**3. Log/API divergence forever.** Internal logs print `RunId::as_str()` = bare ULID. The API returns `run_…`. The whole pitch was "spot the ID and know what it is," but when you're actually debugging in server logs (which is where the hard problems live), you're still looking at bare ULIDs and mentally adding prefixes to correlate with what users reported.

This is duct tape, not a proper approach. The Rust type system already gives compile-time disambiguation between `RunId`, `SnapshotId`, and `GraphCommit` inside the codebase. For external readers, the *field name* (`run_id` vs `graph_commit_id`) already tells you what kind of ID it lives under. The prefix isn't adding new information — it's making the value self-describing in contexts (raw log greps, generic SDK code) where the field-name signal got lost, at the cost of two representations everywhere else.

The "proper" version would push the type tag into storage and into the newtype itself, but that's a real migration with real costs. Doing it at the serialization boundary as a string-mangling step is the worst of both worlds: you pay the complexity tax of two representations without getting the consistency benefits of having one source of truth.

Suggest we close this and let the type system + field names do the disambiguation work they're already designed to do.

ragnorc:
Closing on review feedback. The critique is right: this design locks us into three permanent costs without solving the underlying problem.

1. **Two representations forever.** `01K…` (storage, logs, internal Rust) and `cmt_01K…` (API output) both refer to the same commit because inputs accept both. Any consumer that stores an ID and later compares it against one reconstructed elsewhere hits silent equality bugs. Caches, sets, dedup — two keys for the same thing.

2. **Boundary discipline forever.** Every new HTTP handler, every new output struct has to remember to call `unprefix_*` / `prefix_*`. Both sides are `String`; nothing in the type system enforces it. The first missed call ships an inconsistent response or silently accepts the wrong ID type.

3. **Log/API divergence forever.** Server tracing prints `RunId::as_str()` = bare ULID; the API returns `run_…`. The whole pitch was "spot the ID and know what it is," but when debugging in server logs (where the hard problems live) you're still looking at bare ULIDs and mentally adding prefixes to correlate with what users reported.

The Rust type system already gives compile-time disambiguation between `RunId`, `SnapshotId`, and `GraphCommit` inside the codebase. For API consumers, the field name (`run_id` vs `graph_commit_id`) carries the same information at zero cost.

The "proper" version pushes the type tag into storage and into the newtype itself — a real migration with real cost. Doing it at the serialization boundary as string-mangling is the worst of both worlds.

Closing. If we revisit later it should be the structured-newtype migration, not this.

PR #73, opened 2026-05-04
add schema level embedding model selection

pablocpz:


aaltshuler:
Thanks for the contribution, @pablocpz. Closing this one out — the diff is dominated by unrelated rustfmt changes across ~40 files, the highest-risk paths (schema_apply rewrite, loader auto-embed, query-time model lookup) ship without test coverage, and the embedding calls are per-row sequential which would hold `__schema_apply_lock__` for hours on any non-trivial table. The shape of the feature overlaps with work we're planning to sequence ourselves, so we'll take it from here rather than iterate in this PR.

PR #98, opened 2026-05-15
engine: fix ambiguous-merge false-positive on repeat merge loads

joanreyero:
## Summary

Fixes a Lance `Ambiguous merge inserts are prohibited: multiple source rows match the same target row on (id = ...)` false-positive that fires on the second sequential `omnigraph load --mode merge` against the same keys, even when every source batch has exactly one row per key.

## Root cause

`prepare_updates_for_commit` (`crates/omnigraph/src/db/omnigraph/table_ops.rs:576-599`) eagerly builds a BTREE on `id` after every load. After a subsequent `stage_merge_insert` rewrites rows, the index is never refreshed — its entries still address tombstoned-by-DV slots in the original fragment, while the live rewrites live in a new unindexed fragment.

On the next merge, Lance picks the indexed-scan join path (`create_indexed_scan_joined_stream` in `lance-4.0.0/src/dataset/write/merge_insert.rs:616+`) which UNIONs:
1. an index lookup → returns row_addrs pointing at the tombstoned slots
2. a full scan of `unindexed_fragments` → returns the live rewrites

Both halves surface the same target `_rowid` (Lance preserves stable row ids across update), so the join emits two matched rows with the same row_id, and `MergeState::process_row_action` (`lance-4.0.0/src/dataset/write/merge_insert/exec/write.rs:117`) trips its source-dedup check.

## Diagnostic trail

1. **Reproduced at the engine level** via the regression test in `tests/consistency.rs` — same failure as the bug-report bash repro.
2. **Verified the dataset itself was clean** after the first merge: 55 unique ids, 55 unique `_rowid`s, 55 unique `_rowaddr`s, `TEST-1` mapping to a single `(rowid, rowaddr)` pair. So neither source nor target has any real duplicate — the error is misleading.
3. **Built a Lance-only baseline** (init empty → truncate → append seed → 2× `MergeInsertBuilder::execute_uncommitted` + `CommitBuilder::execute`, no OmniGraph in the loop). Both merges succeeded. So the bug is not in Lance's merge_insert itself, not in the two-phase commit pattern, not in truncate+append.
4. **Compared commit counts.** OmniGraph reaches version 5 after `init + overwrite`; the Lance-only mirror reaches version 3. The 2-commit gap is the BTREE on `id` and INVERTED FTS on `key`, both built by `prepare_updates_for_commit`.
5. **Confirmed hypothesis** by adding a single `create_index_builder(["id"], BTree)` call to the Lance-only baseline before the merges — the bug reproduced exactly.

## Fix

Pass `.use_index(false)` to `MergeInsertBuilder` in both call sites in `crates/omnigraph/src/table_store.rs` (`merge_insert_batch` and `stage_merge_insert`). Forces Lance to take `create_full_table_joined_stream`, which scans the dataset once and respects deletion vectors, avoiding the stale-index / unindexed-fragment union entirely.

The trade-off is a full-table join scan on every merge_insert. This is the only correct option until either:
- the eager scalar-index build is removed from the load commit path (deferring to `ensure_indices` per the "indexes are derived state" invariant in [`docs/dev/invariants.md`](https://github.com/ModernRelay/omnigraph/blob/main/docs/dev/invariants.md)), or
- Lance's indexed-merge path learns to respect deletion vectors.

## Commits

- `tests: regression for repeated merge load against overlapping keys` — red baseline; reviewer can `git checkout` this commit alone and reproduce the failure via `cargo test -p omnigraph-engine --test consistency load_merge_repeated_against_overlapping_keys_succeeds`.
- `engine: disable indexed merge_insert path to avoid stale-index dupes` — fix that turns the test green.

## Test plan

- [x] `cargo test -p omnigraph-engine --test consistency load_merge_repeated_against_overlapping_keys_succeeds` passes
- [x] `cargo test --workspace --locked` passes (~800 tests, no regressions)
- [x] End-to-end repro against RustFS S3 using the exact bash script from the bug report (`init` → `load --mode overwrite` 500 rows → 2× `load --mode merge` of 60 rows) — both merges exit 0

---

<a href="https://app.devin.ai/review/modernrelay/omnigraph/pull/98" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


ragnorc (approved):

aaltshuler:
Superseded by #109 (merged as e1b40ae), which fixes the same Lance dup-rowid `MergeInsertBuilder` bug class via `SourceDedupeBehavior::FirstSeen` instead of `use_index(false)`.

Both approaches were cross-validated on this branch and each individually covers both surfaces (load + mutate). `FirstSeen` was chosen because it has no perf cost — `use_index(false)` forces full-table scans on every `merge_insert`, while `FirstSeen` leaves Lance's indexed-merge path intact and only changes dedup behavior when the buggy code path is hit.

Upstream tracking: lance-format/lance#6877. Window-2 canary test `consistency::load_merge_window_2_documents_upstream_lance_gap` signals when the workaround can be dropped.

Thanks for the report and the parallel investigation — your reproducer on the load surface was load-bearing in nailing the root cause.

PR #211, opened 2026-06-13
refactor(cli): collapse write/query forks onto GraphClient (RFC-009 Phase 3b)

aaltshuler:
Phase 3a put the `GraphClient` enum in place and collapsed the five uniform read forks. **3b folds the remaining data-plane forks onto the same enum**: `load`, `ingest`, `mutate`, `query`, `branch create/delete/merge`, and `schema apply`. The 15 per-command `if graph.is_remote {…} else {…}` forks are now gone.

## Design

- **`Embedded` grew an optional policy context** — `{ uri, graph: Option<ResolvedCliGraph>, actor }`. `resolve()` (reads + `query`) leaves it empty → bare `Omnigraph::open`; `resolve_with_policy()` (writes) fills it → `open_local_db_with_policy` + resolved actor. One `open_embedded` helper picks the path, preserving today's two behaviors exactly.
- **`apply_schema` takes the catalog validator as `impl FnOnce(&Catalog)`** — embedded runs it inside `apply_schema_as_with_catalog_check`, remote ignores it (server runs its own check). That non-object-safe closure is why `GraphClient` is an enum, not a trait. The stored-query registry is still built caller-side, and only for the local path.
- **`load` and `ingest` stay separate methods** — same operation, different surfaced shape (CLI `LoadOutput` with two distinct per-arm mappings vs the wire `IngestOutput`), so no drift risk.
- The now-fully-dead `execute_read`/`execute_read_remote` and `execute_change`/`execute_change_remote` pairs are retired (`legacy_change_request_body` stays — `client.rs` uses it). The export pair remains for 3c.

## Verification

- **`parity_matrix.rs` textually unchanged** (not in the diff) — all 11 rows green.
- `cli_data` / `cli_schema_config` / `cli_queries` green.
- Full `cargo test --workspace --locked` → exit 0, zero failures.

Out of scope (→ 3c): export (streaming), graphs-list, export helper retirement.

### DIFF
--- a/crates/omnigraph-cli/src/client.rs
+++ b/crates/omnigraph-cli/src/client.rs
@@ -4,38 +4,56 @@
 //! 15 per-command `if graph.is_remote { … } else { … }` forks collapse
 //! into two arms here.
 //!
-//! Phase 3a scope: the factory + the uniform read verbs (snapshot,
-//! schema show, branch list, commit list/show — all of which open the
-//! local engine WITHOUT policy today, preserved exactly). Write verbs
-//! and the policy-bearing `query`/`mutate` arrive in 3b (the Embedded
-//! variant will grow the policy context then); export + graphs-list in
-//! 3c. Behavior is unchanged per verb — the Phase-1 parity matrix is the
-//! referee and stays textually unchanged.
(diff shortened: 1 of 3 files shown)


ragnorc:
@greptile review

PR #311, opened 2026-06-28
feat(engine): retire commit-graph tables

ragnorc:
## Summary

Retires the legacy commit-graph Lance datasets and makes `__manifest` the single source of truth for graph lineage, branch heads, and internal storage-version gating.

This PR intentionally changes the upgrade model: this binary serves exactly the current internal manifest schema (`v4`). Graphs written by older releases are refused on open with an explicit export/import rebuild path instead of being migrated in place.

### Commit Graph Retirement

- Deletes the vestigial `_graph_commits.lance` and `_graph_commit_actors.lance` storage paths.
- Makes `CommitGraph` a pure in-memory projection of `__manifest` `graph_commit` / `graph_head` rows.
- Routes branch create/delete authority through `__manifest` only.
- Removes commit-graph table opens, optimization, cleanup, instrumentation counters, old readers, and migration fixtures.

### Strict Internal-Schema Versioning

- Sets `MIN_SUPPORTED == CURRENT == 4`.
- Stamps newly initialized graphs with the current internal schema version.
- Refuses both older and newer manifest stamps at every open/publish boundary.
- Replaces the previous in-place migration dispatcher with a small stamp/refuse gate and actionable error text.

### Operator Visibility

- Surfaces `internal_schema_version` in snapshot output and API response types.
- Adds internal-schema visibility to the CLI/server surfaces covered by this branch.
- Updates OpenAPI fixtures and related API tests.

### Docs

- Adds the export/import upgrade guide.
- Documents strict single-version storage compatibility.
- Updates developer invariants, Lance notes, release notes, constants, maintenance docs, and write-path docs to match the retired commit-graph tables.

## Breaking Change

Graphs created by older releases with a sub-v4 internal schema are no longer opened by this binary.

Upgrade path:

1. Run `omnigraph export` with the older release that can still read the graph.
2. Create a new graph with this release using `omnigraph init`.
3. Import the exported JSONL with `omnigraph load`.

Data, vectors, and blobs are preserved by the export/import rebuild. Commit history and branches are not preserved; export is snapshot-based, so rebuild each branch snapshot intentionally if needed.

## Compatibility Test Coverage

The CLI system test now exercises the full release-to-current lifecycle using published GitHub release binaries:

- `v0.7.2`
- `v0.7.1`
- `v0.7.0`
- `v0.6.2`

For each release, the test:

1. Downloads the platform release tarball and verifies its `.sha256`.
2. Uses that legacy binary to initialize and load a graph.
3. Creates a branch and writes branch-only data.
4. Verifies the current v4 binary refuses the legacy graph with the export/import nudge.
5. Exports the branch snapshot with the legacy binary.
6. Recreates the graph with the current binary.
7. Loads the export and verifies snapshot row counts plus query behavior.

## Validation

- `cargo test -p omnigraph-cli --test system_local local_cli_legacy_release_graphs_refuse_then_export_import_into_v4 -- --nocapture`
  - Passed: `1 passed; 0 failed; 28 filtered out`
- `git diff --check -- crates/omnigraph-cli/tests/system_local.rs`
  - Passed

Earlier branch validation also covered the focused manifest refusal tests, lineage projection tests, CLI snapshot/version tests, and docs link checks.


ragnorc:
## Resolution of the branch-stamp review comments (cursor[bot] / chatgpt-codex / greptile)

The reviewers flagged that the open path validates only **main's** internal-schema stamp, so a branch read could decode a branch stamped outside this binary's range. After weighing it, the resolution is to **gate the stamp at the graph (main) level**, not per-branch, and record the residual as a documented known gap rather than add per-branch machinery.

**Why graph-level is the lower-liability choice:**

- The internal-schema stamp describes the graph's **storage format** — a graph-wide property. It lives per-branch only as an artifact of sitting in `__manifest`'s schema metadata (a branched Lance dataset). The upgrade contract is graph-wide too: the export/import rebuild is whole-graph, and you can't merge across versions.
- With one binary version, every branch is always CURRENT (init stamps main, `create_branch` forks the stamp, the publisher writes rows without re-stamping). A branch stamped out of range while main stays in range is only reachable with concurrent **multi-version writers** — a topology the strand model already declares unsupported ("this binary reads exactly ONE internal-schema version"; multi-process writers are documented one-winner-CAS territory).
- A per-branch read gate would add a second `__manifest` open on every non-main branch read (regressing the RFC-013 read-latency work and tripping the `warm_read_cost.rs` budgets) to defend a state we don't support — unearned complexity.
- The dangerous direction (reading a *future* format) stays partly covered without it: the publisher already refuses *writes* to a future-stamped branch (it reads its target's stamp), and a graph whose main was advanced by a newer binary is refused at open. The residual is a read-only hole in an unsupported topology.

**What changed:** `refuse_if_internal_schema_unsupported` stays single-arg (main), the per-branch refuse and its test were reverted, the greptile message-wording fix was kept, and the read-only residual is now a documented known gap in `docs/dev/invariants.md` (to close only when multi-version write topologies become supported — folded into the branch-manifest open the read already does, at zero extra I/O). See commit `ddd194a3`.

This branch also now retires the dead `commit_graph_reads` cost counter (`24d8eaf2`) and aligns the always-loaded + user-facing docs with the retirement and the strand storage-versioning model, including a new `docs/user/operations/upgrade.md` (export/import recipe) and `docs/dev/versioning.md` (the four-axis compatibility policy) (`05eafc6e`).

PR #342, opened 2026-07-10
Field-level updates: stage only assigned columns in update mutations

ragnorc:
## What & why

An \`update\` that is the only statement touching its type now stages a partial merge source — the row key, the assigned columns, and any columns completing a touched \`@unique\` group — as a matched-only merge, which Lance patches in place on the same fragment. Unassigned columns (including large embedding vectors) are neither read nor rewritten, and indexes over them keep their fragment coverage, closing the class where every update eroded all of a table's index coverage until the next \`optimize\`. Semantics are unchanged; queries mixing an update with other same-table statements fall back to whole-row staging. One pinned residual: the join-key (\`id\`) index still loses patched fragments because the merge source must carry the key — parity with the previous behavior, tripwired to tighten once upstream excludes ON columns from column patches.

## Backing issue / RFC

- [ ] Fixes an **accepted** issue: Closes #
- [ ] Implements / is an **accepted** RFC: <link>
- [ ] **Trivial fast-lane** — no issue/RFC required

Maintainer change (internal process applies); design and validation notes live in the commit messages.

## Checklist

- [x] Tests added/updated for behavior changes (red→green regression tests, fallback pins, a raw-Lance surface guard)
- [x] Public docs updated if user-facing surface changed (docs/user/mutations, docs/dev/writes)
- [x] Reviewed against docs/dev/invariants.md — one commit boundary per table preserved via explicit fallbacks; validation stays loud (composite-unique completion columns); no deny-list item hit

## Notes for reviewers

The eligibility rule is deliberately narrow (partial staging only when the update is the table's sole op in the query): partial and full batches cannot share one uniform-schema merge source, and a present column's null cell means "set NULL", so widening a partial batch would silently null-overwrite — the fallback matrix is pinned by tests. The surface guard pins the Lance contract this rests on (in-place column patch, missing columns untouched, field-scoped index pruning) and both tripwires go red if upstream behavior shifts.

---

> [!NOTE]
> **High Risk**
> Changes core mutation staging, Lance merge semantics, and uniqueness validation on the write path; incorrect partial projection could corrupt rows or miss constraint violations, though eligibility is narrow and heavily tested.
> 
> **Overview**
> Eligible node **updates** (the only statement touching that type in the query, and no `Blob` columns on the type) now stage a **partial** Lance merge: source columns are `id` plus assigned fields only, with **`WhenNotMatched::DoNothing`**, so unassigned columns and their indexes are not rewritten wholesale.
> 
> **`partial_update_plan`** computes assigned columns, transitively completes overlapping `@unique` / `@key` groups for validation, scans only `id` + completion columns, and keeps completion fields on the change-set but **projects them out** before merge staging so Lance does not patch or prune indexes on untouched columns. Mixed same-table ops, chained updates, and blob tables still use **whole-row** merge staging; composite `@key` assignment is rejected for all key columns.
> 
> Staging adds **`PendingMode::PartialUpdate`**, **`append_partial_update_batch`**, and merge instrumentation (**`MergeShape`**: column list + insert disposition). Unique validation **skips** constraint groups with no column present in a partial batch. The schema parser now rejects **`@unique` on Blob/Vector** body constraints (with tests).
> 
> Tests and docs cover partial vs fallback shapes, composite-unique completion, index coverage, Lance partial-merge guards, and user-facing “updates write only assigned columns” behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f30bc18ecd269a646f1d5617ddee6add4d440c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<h3>Greptile Summary</h3>

This PR adds field-level staging for eligible update mutations. The main changes are:

- Partial update planning for `id`, assigned columns, and constraint-completion columns.
- Matched-only partial merge staging with validation-only columns projected out.
- Unique validation support for partial change batches.
- Parser rejection for `@unique` constraints on Blob and Vector fields.
- Tests and docs for index coverage, composite unique validation, and Lance partial-merge behavior.

This looks safe to merge.

- No blocking issues found in the changed code.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| crates/omnigraph/src/exec/mutation.rs | Adds partial update eligibility, column planning, scan shaping, and fallback to full-row staging for mixed or blob-bearing cases. |
| crates/omnigraph/src/exec/staging.rs | Adds partial update pending mode, prevents partial/full batch mixing, and stages matched-only merges after removing validation-only columns. |
| crates/omnigraph/src/validate.rs | Updates unique validation so untouched constraint groups are skipped while partially present groups still fail loudly. |
| crates/omnigraph-compiler/src/schema/parser.rs | Rejects body-level `@unique` constraints that reference Blob or Vector properties. |
| crates/omnigraph/src/table_store.rs | Records merge source columns and unmatched-row behavior for write probes after successful merge staging. |
| crates/omnigraph/src/instrumentation.rs | Adds merge shape probe data used to assert partial staging behavior in tests. |

<h3>Flowchart</h3>

<a href="#gh-light-mode-only">

```mermaid
flowchart TD
  A[Update mutation] --> B{Only op touching this type?}
  B -- No --> F[Whole-row staging]
  B -- Yes --> C{Blob-bearing table?}
  C -- Yes --> F
  C -- No --> D[Build partial plan]
  D --> E[Scan id plus completion columns]
  E --> G[Apply assigned literals]
  G --> H[Validate assigned plus completion columns]
  H --> I[Project validation-only columns out]
  I --> J[Matched-only Lance merge]
  J --> K[Unassigned columns stay untouched]
  F --> L[Existing full-schema merge path]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
  A[Update mutation] --> B{Only op touching this type?}
  B -- No --> F[Whole-row staging]
  B -- Yes --> C{Blob-bearing table?}
  C -- Yes --> F
  C -- No --> D[Build partial plan]
  D --> E[Scan id plus completion columns]
  E --> G[Apply assigned literals]
  G --> H[Validate assigned plus completion columns]
  H --> I[Project validation-only columns out]
  I --> J[Matched-only Lance merge]
  J --> K[Unassigned columns stay untouched]
  F --> L[Existing full-schema merge path]
```

</a>

<sub>Reviews (8): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/modernrelay/omnigraph/commit/7f30bc18ecd269a646f1d5617ddee6add4d440c2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=43332411)</sub>

**Context used:**

- Context used - AGENTS.md ([source](https://app.greptile.com/modernrelay/github/ModernRelay/omnigraph/-/custom-context?memory=59169be6-01cf-4bae-80fc-604b6a708098))
- Context used - CLAUDE.md ([source](https://app.greptile.com/modernrelay/github/ModernRelay/omnigraph/-/custom-context?memory=fba7c581-edb6-48b3-90ce-1db695f47e8b))


crates/omnigraph/src/exec/mutation.rs
```diff
@@ -397,6 +397,74 @@ fn predicate_to_sql(
...
+/// keeps its coverage (Lance prunes only `fields_modified`).
+fn partial_update_plan(
+    node_type: &omnigraph_compiler::catalog::NodeType,
+    full_schema: &SchemaRef,
+    assignments: &[IRAssignment],
+) -> PartialUpdatePlan {
+    let assigned: HashSet<&str> = assignments.iter().map(|a| a.property.as_str()).collect();
+    let mut completion: HashSet<&str> = HashSet::new();
+    for group in &node_type.unique_constraints {
+        if group.iter().any(|col| assigned.contains(col.as_str())) {
```
aaltshuler: **Correctness — overlapping `@unique` groups make a legal sole update hard-fail (verified).** This loop pulls in only groups intersecting the *assigned* set — no transitive closure — while `evaluate_unique` (validate.rs) hard-errors on a *partially present* group.

Repro: `@unique(room, hour)` + `@unique(hour, day)`, query `update T set { room: $r } where …` → completion = {room, hour}; batch = (id, room, hour); group `(hour, day)` is partially present (`hour` yes, `day` no) → the whole mutation fails with `missing unique column 'day'` though the update touches neither column. Nothing rejects overlapping groups (the parser only checks each column exists), and multiple `@unique(...)` bodies are grammatical.

Secondary trigger: a composite `@key` sharing a column with a `@unique` group — `constraints_for` registers the `@key` group as a Unique constraint too, but this fn iterates only `node_type.unique_constraints`, never `node_type.key`.

Fix shape: fixed-point-close the completion set over transitively overlapping groups (including the key group), or pass the assigned set into the evaluator so it skips any group the write doesn't intersect.

aaltshuler (commented):
Reviewed the diff plus the enclosing code at `985182d8`, with every substrate claim re-verified against the pinned Lance 9.0.0-beta.15 checkout. The core mechanism is sound and well-defended: the surface guard exercises the exact production path (BTREE on the join key → legacy Merger, which at our pin locates keys **by name**, so there is no column-order hazard), change-feed visibility and read-your-writes were traced clean, and `InsertAll→DoNothing` is safe under the eligibility rule. Two correctness findings survive verification (one reproduced end-to-end at both commits), plus robustness/cleanup items — all inline below.

One finding with no inline anchor: **`docs/dev/testing.md` is not updated in this PR** (maintenance rule 1 — "update both the source code and the doc in the same change"). Four test files gained significant coverage (staging-shape tests in `writes.rs`, two index-coverage cells in `scalar_indexes.rs`, the composite-unique cell in `validators.rs`, the partial-merge surface guard) and the always-on test map describes none of it — the drift class `5c3125e7` recently existed to fix.

Verified and deliberately **not** flagged, to save you the re-derivation: the "partial staging rides the legacy indexed Merger" hazard is real as a path fact but harmless — at 9.0.0-beta.15 `extract_selections` locates join keys by name (the positional check was itself the lancedb#3515 bug, since fixed), and the id-first ordering is not load-bearing; the `evaluate_unique` all-absent skip cannot hide a real violation for blob-bearing groups, because `unique_key_scalar` makes a non-null blob under `@unique` un-committable on every write surface.

crates/omnigraph/src/exec/mutation.rs
```diff
@@ -397,6 +397,74 @@ fn predicate_to_sql(
...
+    let mut stage_cols = Vec::new();
+    for field in full_schema.fields() {
+        let name = field.name().as_str();
+        let is_assigned = assigned.contains(name);
+        let in_completion = completion.contains(name);
+        if name == "id" || is_assigned || in_completion {
+            output_fields.push(field.clone());
+        }
+        if name == "id" || is_assigned {
+            stage_cols.push(name.to_string());
```
aaltshuler: **Correctness — the partial plan is Blob-blind; one mode reproduced end-to-end.**

(a) *Reproduced at both commits:* `update Document set { content: $c }` with param `c = null` on a `content: Blob?` property. `apply_assignments` omits a blob column whose resolved value isn't `Literal::String`, but `stage_cols` here unconditionally lists every assigned column → staging fails with `manifest_internal("partial-update stage column 'content' missing from accumulated batch…")`. At the merge-base the same query **succeeds** (blob left untouched) — and it still succeeds post-PR when another op rides the query (whole-row fallback), so the outcome now depends on unrelated statements in the same query.

(b) *Traced:* `scan_fields` doesn't exclude blobs either, so `@unique(name, data)` with `data: Blob` plus a sole update assigning `name` puts the blob into the scan projection — which the projection comment further down says Lance's scanner asserts on.

Fix: exclude non-String-assigned blob columns from `stage_cols` (mirroring `apply_assignments`' omission rule, or reject null blob assignment as a typed user error on *both* paths) and filter blobs from the scan schema. Cheap insurance for (b): the body-level `@unique` parser arm doesn't reject Blob/Vector the way the `@key`/`@index` arms do (`schema/parser.rs` ~944) — one guard closes that class at compile time.

crates/omnigraph/src/exec/staging.rs
```diff
@@ -52,6 +52,18 @@ pub(crate) struct PendingTable {
...
+    /// completion) columns for the validation change-set, but the STAGED merge
+    /// source is projected down to `stage_cols` = (merge key + assigned) and
+    /// staged matched-only (`WhenNotMatched::DoNothing`). Completion columns
+    /// are validation inputs, never merge inputs: Lance counts every source
+    /// column as modified, so staging an unassigned `@unique`-group member
+    /// would patch it and prune its index for no semantic reason. Set only by
+    /// [`MutationStaging::append_partial_update_batch`], whose eligibility
+    /// rule (the table's ONLY op in this query is a single `update`)
+    /// guarantees no full-schema batch ever lands on the same table.
+    pub(crate) partial_stage_cols: Option<Vec<String>>,
```
aaltshuler: **Robustness — partial-ness is an `Option` side-field rather than part of the mode the code dispatches on.** `stage_pending_table` matches `table.mode` three times (stage-kind, combine/dedupe, stage dispatch); only the last consults `partial_stage_cols`. A future consumer that matches `PendingMode::Merge` — a finalize-coalescing change (exactly what the accumulator's own docs invite), a retry/rebase path, a staged-scan union — treats a partial batch as full-row with no compiler help, and silently widening one produces the null-overwrite this PR's comments correctly identify as unsound. A `PendingMode` variant carrying the partial columns (e.g. `Merge { partial: Option<Vec<String>> }`) forces every match site to disposition it.

crates/omnigraph/src/table_store.rs
```diff
@@ -1177,6 +1177,13 @@ impl TableStore {
             ));
         }
         let merged_rows = batch.num_rows() as u64;
+        let inserts_unmatched = matches!(when_not_matched, WhenNotMatched::InsertAll);
+        let source_columns: Vec<String> = batch
```
aaltshuler: **Efficiency — per-merge `Vec<String>` of all source column names allocated in production for a test-only probe.** Every `stage_merge_insert` — every mutation query and every load — now pays a per-column `String` clone before the merge, dropped unused when no probes are installed. This breaks the instrumentation module's no-op-in-production pattern (cf. `record_open`, which passes a cheap borrow and classifies *inside* the `try_with` closure). Fix: capture `batch.schema()` (one Arc bump — already done below for the reader) and materialize the name vec inside `record_stage_merge_insert`'s `try_with`.

crates/omnigraph/tests/writes.rs
```diff
@@ -2065,3 +2065,162 @@ async fn filtered_read_after_append_and_delete_is_consistent() {
...
+        "a sole update stages a matched-only merge (WhenNotMatched::DoNothing)"
+    );
+}
+
+/// Fallback pin: a query mixing insert + update on ONE table stages full rows
+/// as an upsert — partial and full batches cannot share one merge source, and
+/// one table commits at most one version per query (invariant 4). Pins the
+/// RFC-022 fallback so the partial path can never split a table's commit.
+#[tokio::test]
+async fn mixed_insert_update_same_table_stages_full_row_upsert() {
```
aaltshuler: **Test duplication (testing.md extend-vs-new rule).** This test and `chained_updates_same_table_stage_full_rows` duplicate areas existing tests already own: `mixed_insert_and_update_on_same_person_coalesces_to_one_merge` (line 674) already pins insert+update-on-one-table coalescing — and its `STAGED_QUERIES::insert_then_update_same_person` fixture is the exact query shape `MIXED_INSERT_UPDATE` re-mints — while `chained_updates_with_overlapping_predicate_respects_intermediate_value` (line 1080) owns the chained-update case via `update_then_filter_by_old_value`. Wrapping those tests' `mutate` calls in `with_merge_write_probes` and adding the shape assertions there pins the fallback in one place; two copies of fallback behavior will drift.

crates/omnigraph/src/exec/mutation.rs
```diff
@@ -916,6 +984,29 @@ impl Omnigraph {
...
     ) -> Result<MutationResult> {
+        // RFC-022 eligibility: a node table whose ONLY op in this query is a
+        // single `update` stages a partial-schema matched-only merge (key +
+        // assigned + constraint-completion columns). Every other combination
+        // falls back to whole-row staging: partial and full batches cannot
+        // share one uniform-schema merge source (a present column's null cell
+        // means "set NULL", so widening a partial batch would null-overwrite),
+        // a second same-table op's read-your-writes scan must see full rows,
+        // and one table commits at most one version per query (invariant 4).
+        let mut update_op_census: std::collections::HashMap<&str, (usize, usize)> =
```
aaltshuler: **Simplification — the census is over-built and runs for every mutation query.** At the `Update` arm, `ops == 1` already implies `updates == 1` (the census counts the current op itself), so the second counter and the `is_update` flag are dead weight; the `HashMap` is allocated even for insert-only/delete-only queries (the common bulk shapes); and on the partial path `non_blob_cols` below is still computed then discarded. An inline count at the Update arm — `ir.ops.iter().filter(|op| touches(op, type_name)).count() == 1` — replaces the whole pre-pass (op lists are tiny), and `non_blob_cols` can move into the projection's `else` arm.

ragnorc:
Review dispositions (all six findings confirmed valid; fixes in the latest commit, both correctness items reproduced red-first with the exact predicted errors):

1. **Overlapping `@unique` groups (correctness)** — confirmed and fixed. Completion is now the fixed-point closure over every `@unique` group **and** the `@key` group; the repro (`@unique(room, hour)` + `@unique(hour, day)`, sole update assigning `room`) is a passing regression test, and a genuine transitive collision is still rejected. The composite-`@key` secondary is also closed at the source: key-assignment rejection now checks every key column, not just the first.
2. **Blob-blind plan (correctness)** — confirmed and fixed. The staged source is now derived by *subtraction* (batch schema minus validation-only columns), so `apply_assignments`' omissions can never desync it — null-to-optional-Blob is the historical no-op again on both paths. Blobs are excluded from completion/scan projection, and the parser now rejects `@unique` on Blob/Vector at compile time (mirroring the `@key`/`@index` guards), closing (b) as a class.
3. **Partial-as-side-field (robustness)** — adopted: `PendingMode::PartialUpdate` is a distinct variant, so every match on the mode must disposition partial batches explicitly; the columns remain data on the entry with an internal-error guard on the invariant.
4. **Probe allocation (efficiency)** — adopted: the schema arrives as a borrowed `Arc` and column names materialize only inside the `try_with` closure (the `record_open` pattern); zero production allocations.
5. **Test duplication** — adopted: the fallback-shape assertions moved into the existing owner tests (mixed insert+update coalesce; chained updates), and the two duplicates plus their fixture queries are deleted.
6. **Census simplification** — adopted: inline count at the update arm; the per-query map and the discarded `non_blob_cols` computation are gone.

Suites green: engine (writes/validators/scalar_indexes/end_to_end) + compiler (incl. the two new parser guards).

PR #369, opened 2026-07-18
Make branch delete's dependency check manifest-only per surviving branch

ragnorc:
## What & why

`branch_delete`'s dependency check consumes only each surviving branch's manifest `table_branch` entries, but resolved them through the full `snapshot_of` path — per foreign branch that re-reads the schema contract (3 storage GETs), re-compiles and re-validates the schema source, and assembles the resolved-target projection, O(branches) redundant work on every delete on top of the per-branch manifest scan. This switches the loop to the existing manifest-only per-branch snapshot (the same shape `cleanup` and the write-authority capture use), and adds a cost gate pinning the loop's `__manifest` slope at one read/open per surviving branch so a second per-branch authority read cannot creep back in.

## Backing issue / RFC

- [ ] Fixes an **accepted** issue: Closes #
- [ ] Implements / is an **accepted** RFC: <link>
- [ ] **Trivial fast-lane** — no issue/RFC required

Maintainer perf fix from a branch-operation cost audit; no pre-existing accepted issue.

## Checklist

- [x] Tests added/updated — new `branch_control_cost.rs` slope gate on the shared cost harness (measures exactly 1 manifest read + 1 internal open per surviving branch; the historical two-scan cold resolve measured 2 + 2)
- [x] Public docs updated — docs/dev/testing.md row for the new cost test; no user-facing surface change
- [x] Reviewed against docs/dev/invariants.md — no invariant weakened; removes cold re-derivation on a control path

## Notes for reviewers

The dropped per-foreign-branch schema validation was incidental coupling: it could wedge deletion of an unrelated branch behind another branch's schema drift, while the deleting operation's own schema is still validated under the schema gate — the same rationale the write-authority capture documents for using the unchecked manifest-only snapshot. The dependency predicate and refusal messages are unchanged. Note the recent combined state+lineage coordinator decoding already removed the loop's second per-branch `__manifest` scan; this PR removes the remaining per-branch schema-contract reads/recompiles and pins the slope.


aaltshuler:
Closing this PR as stale against the current write path.

Since this branch was cut, the B1 work changed branch-delete admission, recovery, lifecycle fencing, gate ownership, and operation-local capture. The PR's slope test now passes on current `main` without its production change, while a direct rebase adds a cold `main` manifest open/scan and fails the existing branch-control cost budget. The manifest-only loop also drops per-survivor catalog validation and still performs one history-sensitive manifest scan per surviving branch, so the stated complexity result is not established.

The underlying idea may be revisited as a current-main fast-lane optimization after measuring the remaining schema/lineage cost, preserving catalog validation, and avoiding the extra `main` scan.

PR #489, opened 2026-08-11
Supervise configured graph availability, with typed failures

ragnorc:
## Summary

Stacked on #488. The registry retains every *configured* graph — including ones that fail to open — and a per-graph supervisor drives open and recovery. The first commit introduces that; the rest close what it landed with.

**The retry classifier could never return "retryable."** `FailureClass::Io`/`Timeout` came only from `OmniError::Io`, which the engine has no construction site for. Every Lance and object-store failure arrived as an opaque string and landed in `Unknown`, which is not retried — so a transient S3 error pinned a graph permanently and `full_jitter_backoff` plus the whole `Opening{next_retry}` arm were unreachable code. The existing unit test passed only because it hand-built `OmniError::Io` values nothing emits. Now consumes the classification #491 derives at the substrate boundary, and the arm that sniffed error text for `"internal schema v"` is gone — a wire-visible field must not depend on internal prose.

**One failed refresh could permanently kill reads.** The non-retryable branch dropped the live handle and stored `Unavailable`, a terminal state with no way back — and because of the above, *every* recovery failure took it. Recovery is a write concern: a failed `refresh()` leaves the engine's live view intact, `Dataset::restore` appends rather than removes, reads pin an exact version, `cleanup` refuses while a sidecar is pending, and byte-destructive paths are guarded to objects absent from the live manifest. `Serving` is now a one-way door: a permanent failure is `Blocked { retry_at: None }` — writes blocked, reads intact. `Unavailable` is reachable only from `Opening`.

**A newer recovery request could be clobbered.** Attempts read state, refresh without the entry lock, then store a result derived from that read; the success path re-checked only that the graph was still `Serving`, which every write state satisfies. Entries now carry a generation, bumped per trigger and captured after the permit, and an attempt publishes only if it still holds.

**Backoff never escalated.** Coalescing reset `attempts` to zero, so sustained write traffic held the supervisor at its shortest delay indefinitely — each retry a full recovery sweep whose cost grows with commit depth. Coalescing now preserves the attempt count and any scheduled deadline.

**Attempts were unbounded in time and awaited inline**, holding one of four permits. Four graphs stalled on an unresponsive store starved every other graph, and the loop saw neither shutdown nor new notifications while one ran. Wrapping the attempt in a timeout would be wrong — a recovery mutates durable state under an armed sidecar, and dropping that future mid-restore is exactly what the tracked-write executor exists to prevent. So the 240s deadline bounds the **wait**, not the work: attempts are spawned into a `TaskTracker` the set owns, run to a terminal result, and publish under their own generation check. The loop starts no second attempt for a graph while one is in flight.

Also: typed `GraphState` with one conversion to the wire form (replacing a `&'static str` re-parsed by two matches that each silently reported "unavailable" for anything unrecognised); public 503s carry graph id and class only, never backend text, since that response is built in middleware ahead of any per-graph policy check; `Retry-After` only when a retry is actually scheduled; a graph-local invalid URI becomes an unavailable entry instead of aborting the process; one process-global attempt-permit pool; and `SupervisorSet::idle()` is removed so a live write executor can never be paired with a set that has no consumer for its wake-ups.

## Verification

- Classification mapping across all four storage kinds, including that retryable stays retryable.
- `a_trigger_arriving_mid_attempt_is_not_clobbered` — drives the race through the attempt seam.
- `coalescing_preserves_the_backoff_ladder`; `a_permanent_recovery_failure_keeps_reads_available`.
- Public 503 asserts graph id + class and nothing else.
- `cargo test --workspace --locked --features omnigraph-engine/failpoints,omnigraph-cluster/failpoints` — 75 suites, zero failures.
- Workspace Clippy with warnings denied; `cargo fmt --all --check`.

## Known, deliberate

`mark_recovering` still has no production caller *at this commit* — the caller is the tracked-write executor in #490. Reviewers checking this PR in isolation will correctly observe that the recovery half of the state machine has no producer yet; that is a property of the stack split, not a defect.

Cluster-quarantined graphs (recovery sidecars) are still filtered before settings are built, so they 404 rather than 503. Converting them to configured entries is follow-up work.

No storage-format migration.

<h3>Greptile Summary</h3>

The PR retains all configured graphs in the registry and supervises opening and recovery independently while exposing bounded availability metadata.
- Adds atomic graph runtime states, retry classification, jittered backoff, and process-wide attempt limiting.
- Returns structured unavailable-graph responses and configured graph status through the HTTP API.
- Allows default boot with no healthy graphs while preserving strict all-graph startup.

The PR appears safe to merge because the previously reported backend-diagnostic disclosure has been removed and no blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph-server/src/lib.rs | Integrates configured-registry startup and emits unavailable responses using bounded state and failure-class values rather than backend diagnostics. |
| crates/omnigraph-server/src/registry.rs | Introduces immutable configured entries with atomically readable serving, recovery, and unavailable runtime states. |
| crates/omnigraph-server/src/supervisor.rs | Implements bounded per-graph opening and recovery attempts with classification, coalescing, and retry backoff. |
| crates/omnigraph-server/src/handlers.rs | Lists configured graph availability and distinguishes unavailable configured graphs from unknown IDs. |
| crates/omnigraph-api-types/src/lib.rs | Adds shared graph-state and structured unavailable-response DTOs. |
| openapi.json | Updates the published API schema for graph status and unavailable responses. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant Auth
    participant Registry
    participant Supervisor
    participant Engine
    Client->>Auth: "Request /graphs/{id}/..."
    Auth->>Registry: Resolve configured graph
    alt Graph is serving
        Registry-->>Client: Dispatch with live handle
    else Graph is opening or unavailable
        Registry-->>Client: 503 graph_unavailable
    end
    Supervisor->>Engine: Open or refresh
    alt Attempt succeeds
        Engine-->>Supervisor: Healthy handle
        Supervisor->>Registry: Mark ready
    else Retryable failure
        Engine-->>Supervisor: Classified failure
        Supervisor->>Registry: Publish retry metadata
        Supervisor->>Supervisor: Full-jitter backoff
    else Permanent failure
        Engine-->>Supervisor: Non-retryable failure
        Supervisor->>Registry: Mark unavailable/degraded
    end
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(server): give every AppState a live ..."](https://github.com/modernrelay/omnigraph/commit/3fb26d8f78d01790b87a7da0dfb2f3dff4ee5bc8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52410932)</sub>

**Context used:**

- Knowledge Base — [Server Request Handling](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/server-request-handling.md)
- Knowledge Base — [CLI and Config](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/cli-and-config.md)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5180,6 +5180,7 @@ dependencies = [
  "tempfile",
  "thiserror",
  "tokio",
+ "tokio-util",
  "tower",
  "tower-http",
  "tracing",
@@ -7627,6 +7628,7 @@ dependencies = [
  "bytes",
  "futures-core",
  "futures-sink",
+ "futures-util",
  "pin-project-lite",
  "tokio",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5161,6 +5161,7 @@ dependencies = [
  "clap",
  "color-eyre",
  "dashmap",
+ "fastrand",
  "futures",
  "lance",
  "lance-index",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ base64 = "0.22"
 ariadne = "0.4"
 regex = "1"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
+tokio-util = { version = "0.7", features = ["rt"] }
 object_store = { version = "0.13.2", default-features = false, features = ["aws", "fs"] }
 fail = "0.5"
 time = { version = "0.3", features = ["formatting"] }
--- a/crates/omnigraph-api-types/src/lib.rs
+++ b/crates/omnigraph-api-types/src/lib.rs
@@ -665,7 +665,27 @@ pub struct ResourceLimitOutput {
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 pub struct RecoveryRequiredOutput {
+    /// The durable operation that blocked this request. This is not a request
+    /// identifier for the current HTTP call.
     pub operation_id: String,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum GraphState {
+    Ready,
+    Recovering,
+    Degraded,
+    Opening,
+    Unavailable,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct GraphUnavailableOutput {
+    pub graph_id: String,
+    pub state: GraphState,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub retry_after_seconds: Option<u64>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
@@ -697,6 +717,10 @@ pub struct ErrorOutput {
     /// retry. Its table effects may or may not have started.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub recovery_required: Option<RecoveryRequiredOutput>,
+    /// Present when the graph id is configured but cannot currently serve the
+    /// requested operation. Unknown graph ids remain ordinary 404 responses.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub graph_unavailable: Option<GraphUnavailableOutput>,
 }
 
 pub fn snapshot_payload(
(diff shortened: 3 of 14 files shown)


aaltshuler:
## Maintainer review: changes requested

The state machine and generation fencing are promising, but this is not merge-ready:

1. **The 240-second timer does not prevent starvation.** A timed-out open/refresh retains its global semaphore permit until the underlying call returns. Four hung attempts can still block every graph forever. initial_open_all is also unbounded and waits for every graph before the listener binds. Add tests proving one stalled graph cannot block server startup or another graph.
2. **GET /graphs breaks rolling compatibility.** state, read_ready, and write_ready are newly required fields without Serde defaults, so a new CLI cannot deserialize an older server response. GraphState also needs forward-compatible unknown handling. Add exact old-payload coverage.
3. **A read-only handle can report false recovery success.** #488 read-only refresh can return Ok without Full recovery, while this supervisor interprets every Ok as write-ready. Fence writable capability or return a typed refresh outcome.
4. **Recovery has no production trigger until #490.** mark_recovering has no current caller, despite RecoveryRequired advertising Retry-After. Do not ship this half independently.
5. **Lifecycle ownership is incomplete.** The final stack must close write admission, drain tracked writes, stop supervisor producer loops, then bounded-wait owned recovery attempts under one shutdown budget.
6. **Docs/OpenAPI disagree with behavior.** Cluster docs describe zero-open boot and configured-graph 503s, while server docs still describe startup refusal and 404; read routes omit their possible middleware 503.

Current CI is red on Clippy and the canonical workspace/RustFS jobs were skipped. Required order: corrected #491, fixed #488, then rebase/fix this PR, followed immediately by a rebuilt #490.

aaltshuler:
Closing as superseded by the architecture series: RFC-034 in #496 owns durable recovery authority, RFC-035 in #498 owns served-operation lifetime and shutdown, and RFC-036 in #499 owns atomic runtime activation and supervision. The useful availability requirements and tests from this PR remain inputs to RFC-036, but the current same-handle recovery implementation is not being carried forward. PR #491 remains independent and unchanged.

PR #502, opened 2026-08-13
fix(engine): stop init cleanup from deleting a completed graph

### LINKED ISSUE #495: bug: after a failed init, cleanup deletes a completed store's schema files but not its data, leaving it unable to open or re-init
### What happened

If Omnigraph::init fails at its tail, after GraphCoordinator::init has completed, the store on disk is already complete: the three schema files, every Lance per-type dataset, and __manifest/ with its version stamp are durably written. The only remaining work was returning Ok. The failpoint doc at crates/omnigraph/src/db/omnigraph.rs:3325 states exactly this ("all schema files plus Lance per-type datasets and __manifest/ are on disk"). The error path then runs best_effort_cleanup_init_artifacts (db/omnigraph.rs:3367, called from the init error path at :447). It deletes _schema.pg, _schema.ir.json, and __schema_state.json. By its own doc comment it cannot remove Lance datasets or __manifest/ (recursive deletion needs the deferred delete_prefix primitive).

The result is a store with full Lance data and a stamped manifest but no schema files. Omnigraph::open (read-write and read-only) fails with an untyped internal error: "storage read failed for '<root>/_schema.pg': Object at location ... not found". Plain re-init fails on a raw Lance "Dataset already exists". init --force cannot recover either, since force does not purge Lance state by its documented semantics. Only manually deleting the graph root recovers.

The brick is cleanup-induced, not crash-induced. A crash at the same point (failpoint action "panic", so the cleanup never runs) leaves the schema files in place, and the store opens clean and works. Both directions are pinned by the repro below. This window is also untouched by the recent init crash-safety fix (#487): that fix closes the adjacent torn-manifest-stamp brick in the same init tail, and this bug reproduces unchanged after it merged.

The same cleanup is correct at the early init windows: after a failure when only schema files exist, deleting them lets re-init recover cleanly (init_failpoint_after_schema_contract_written_cleans_up_all_schema_files pins that). The defect is that the cleanup has no notion of "the store is already complete". The existing test init_failpoint_after_coordinator_init_cleans_up_schema_files (tests/failpoints.rs:10281) asserts the deletion at this window as expected behavior but never checks whether the store is usable afterward; a fix flips that test.

Reachability: any error returned from init's tail after coordinator init. In-tree today that is validate_schema_ir_against_snapshot (db/omnigraph.rs:3350, immediately after GraphCoordinator::init_with_session completes), and any future code added to the tail inherits the hazard. A transient storage error is the realistic trigger: a backend that is fully down also fails the best-effort deletes (store survives by accident), but a momentary error fails the tail check and then lets the deletes succeed.

Note the brick comes specifically from the deletion being PARTIAL. Skipping the cleanup leaves a working store (best case). Deleting everything would leave an empty root where re-init succeeds (wasteful, but self-healing, and at init time the datasets hold no user data). Only the middle state, schema files gone with data and manifest left behind, can neither open nor re-init. The cleanup's reach is set by which delete operations exist (no delete_prefix), not by which end state makes sense.

Expected: a failure after the point of durable completion must not destroy the store. Fix directions: check viability before cleanup (a present and stamped manifest means the store is complete, so skip the cleanup and return the original error noting the store is intact; since #487 the stamp is written atomically with the manifest's Create commit, so manifest-present-and-stamped is a reliable completeness test), or reorder the tail so nothing failable runs after the completion point. Either way, opening a store whose data and manifest exist but whose schema files are missing should produce a typed diagnosis naming exactly that, not a raw storage read error. All file:line anchors are from main 34341e9.

### Steps to reproduce

1. Save the test below as crates/omnigraph/tests/init_tail_cleanup_repro.rs.

```rust
use fail::FailScenario;
use omnigraph::db::Omnigraph;
use omnigraph::failpoints::{ScopedFailPoint, names};
use serial_test::serial;

const SCHEMA: &str = "node Person { name: String @key }\n";

#[tokio::test]
#[serial]
async fn init_tail_error_cleanup_bricks_completed_store() {
    let _scenario = FailScenario::setup();
    let dir = tempfile::tempdir().unwrap();
    let uri = dir.path().to_str().unwrap();

    {
        let _fp = ScopedFailPoint::new(names::INIT_AFTER_COORDINATOR_INIT, "return");
        let err = match Omnigraph::init(uri, SCHEMA).await {
            Ok(_) => panic!("expected init to fail at the tail failpoint"),
            Err(e) => e,
        };
        println!("init error: {err}");
    }

    println!("_schema.pg exists after cleanup: {}", dir.path().join("_schema.pg").exists());
    let reopen = Omnigraph::open(uri).await;
    println!("reopen: {:?}", reopen.as_ref().err().map(|e| e.to_string()));
    let reinit = Omnigraph::init(uri, SCHEMA).await;
    println!("re-init: {:?}", reinit.as_ref().err().map(|e| e.to_string()));
    assert!(reopen.is_err() && reinit.is_err());
}

#[tokio::test]
#[serial]
async fn contrast_crash_without_cleanup_leaves_store_openable() {
    let _scenario = FailScenario::setup();
    let dir = tempfile::tempdir().unwrap();
    let uri = dir.path().to_str().unwrap().to_string();

    fail::cfg("init.after_coordinator_init", "panic").unwrap();
    let u = uri.clone();
    let join = tokio::spawn(async move { Omnigraph::init(&u, SCHEMA).await.map(|_| ()) }).await;
    fail::remove("init.after_coordinator_init");
    println!("init crashed, cleanup never ran: {}", join.is_err());

    println!("_schema.pg exists after crash: {}", dir.path().join("_schema.pg").exists());
    assert!(Omnigraph::open(&uri).await.is_ok(), "store must open after crash without cleanup");
}
```

2. Run: cargo test -p omnigraph-engine --features failpoints --test init_tail_cleanup_repro -- --nocapture --test-threads=1

3. Observed: the first test's init fails with the injected error, _schema.pg is gone, reopen fails with the raw "storage read failed for '<root>/_schema.pg'" error, and re-init fails with the raw Lance "Dataset already exists" error. The second test's init dies by panic at the same window, the schema files survive, and the store opens cleanly.

Expected: after an init failure at a point where the store is durably complete, the store either opens normally (the error return notes it is intact) or fails with a typed diagnosis naming the missing schema files; re-init on the same root is either possible or refused with a truthful message. The contrast test documents that the store is viable whenever the cleanup does not run.

Narrow-trigger note: only the error-return path at the tail window triggers the brick. Failures at the early windows (after_schema_pg_written, after_schema_contract_written) recover cleanly via re-init, and a hard crash at the tail window is harmless. The failpoint stands in for any real error in init's tail, such as validate_schema_ir_against_snapshot failing.

### Version

omnigraph 0.10.0, main @ 34341e9, built from source (Lance 10.0.0)

### Environment

macOS (Darwin 25.5, arm64), local file:// store. Storage-independent: the cleanup is plain storage deletes and the failpoint is in the engine's init path, not the storage layer. Also reproduced on an in-memory object-store backend under the harness.

### Logs / output

```shell
init error (tail failure after the store is complete): injected failpoint triggered: init.after_coordinator_init
_schema.pg exists after cleanup: false
reopen: FAILED: storage read failed for '/tmp/.tmpM6x3Oo/_schema.pg': Object at location /tmp/.tmpM6x3Oo/_schema.pg not found: No such file or directory (os error 2)
re-init: FAILED: storage: Dataset already exists: /tmp/.tmpM6x3Oo/nodes/0000000000000001-0000000000000003, lance-10.0.0/src/dataset/write/insert.rs:295:28

init crashed (panic, cleanup never ran): true
_schema.pg exists after crash: true
reopen after crash-without-cleanup: OK
```

azimafroozeh:
## What & why

Closes #495. `Omnigraph::init` runs its best-effort schema cleanup on every error, including errors returned after `GraphCoordinator::init` has committed the stamped `__manifest`, which is the point where the graph is durably complete. Past that point the cleanup deletes the three schema files while the Lance data stays, leaving a graph that neither opens (raw storage read error) nor re-inits (`Dataset already exists`); the only recovery is deleting the root by hand. A process crash at the same point leaves an openable graph, so the damage is entirely cleanup-induced: backward recovery applied past the commit point.

- Split `init` at the commit point: the commit phase now ends at the `__manifest` commit, so an error from it is provably pre-commit and keeps today's cleanup; the post-commit checks run as a separate step whose errors return with the graph untouched and a message stating the graph is intact and opens normally.
- Open-path floor for already-damaged graphs: when `__manifest` and data are present but `_schema.pg` is missing, `open` returns a typed not-found naming the state and the remedy (restore the schema files, or rebuild via export) instead of a raw storage read error.
- Both error surfaces change intentionally and are pinned by tests: the init tail error gains the intact-graph note, and the open path trades an untyped storage error for the typed diagnosis. Callers matching on the old texts must update.

## Future work

A forward-recovery path that completes a post-commit failure instead of reporting it was considered and deferred; it needs the recovery-sidecar protocol extended to init and is worth its own RFC.

## Local verification

- `cargo test -p omnigraph-engine --features failpoints --test failpoints`: green; the flipped pinning test `init_failpoint_after_coordinator_init_leaves_completed_store_intact` failed pre-fix with the predicted signature (schema files deleted, raw storage read error on reopen), and the new `open_missing_schema_pg_reports_schema_files_missing` covers the floor.
- `cargo test --workspace --locked --features omnigraph-engine/failpoints,omnigraph-cluster/failpoints`: green, plus `scripts/check-agents-md.sh` after the docs edit.
- `cargo clippy --workspace --all-targets --locked -- -D warnings -W clippy::dbg_macro` on both CI feature graphs and `cargo fmt --all --check`: clean.

<h3>Greptile Summary</h3>

The PR separates initialization at the durable manifest commit boundary and adds exact-attempt recovery for acknowledgement-unknown creates.
- Preserves schema artifacts whenever physical initialization may have committed.
- Wraps confirmed post-commit failures in typed initialization errors.
- Adds a durable root-scoped initialization claim, exact-genesis probing, damaged-schema diagnostics, and failpoint coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph/src/db/omnigraph.rs | Splits initialization into pre-physical, commit, ambiguous-outcome, and post-commit phases while ensuring only pre-physical failures authorize schema cleanup. |
| crates/omnigraph/src/db/manifest/graph.rs | Adds attempt-local genesis receipts, explicit manifest-create outcome classification, exact-genesis authentication, and a separate post-commit state load. |
| crates/omnigraph/src/db/graph_coordinator.rs | Exposes distinct coordinator operations for manifest commit, exact-genesis recovery, and post-commit assembly. |
| crates/omnigraph/src/error.rs | Adds typed claimed, committed, and indeterminate initialization outcomes for callers and operators. |
| crates/omnigraph/tests/failpoints.rs | Exercises confirmed post-commit failures, lost acknowledgements, failed probes, cleanup failures, retained claims, and successful reopening. |
| crates/omnigraph/src/db/manifest/tests.rs | Verifies that exact-genesis probing accepts only the receipt minted by the creating initialization attempt. |
| crates/omnigraph/src/db/omnigraph.rs | The earlier cleanup-boundary and unannotated-error findings are both resolved at current HEAD. |

</details>

```mermaid
flowchart TD
  A[Acquire durable init claim] --> B[Write schema contract]
  B --> C[Create physical tables and manifest]
  C -->|Confirmed manifest commit| D[Run post-commit checks]
  C -->|Outcome unknown| E[Probe exact genesis receipt]
  E -->|Exact receipt found| F[Finish coordinator validation]
  E -->|Probe fails or mismatches| G[Return InitializationIndeterminate and retain claim]
  D -->|Success| H[Release claim and return graph]
  D -->|Failure| I[Release claim and return InitializationCommitted]
  F -->|Success| H
  F -->|Failure| I
  B -->|Failure before physical init| J[Clean owned schema artifacts and release claim]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(engine): make graph initialization o..."](https://github.com/modernrelay/omnigraph/commit/79a0d2e5eecb91671c31a2e65e136321537e2e98) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53131382)</sub>

- Knowledge Base — [OmniGraph write path](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/omnigraph-write-path.md)
- Knowledge Base — [Dev conventions and invariants](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/dev-conventions-and-invariants.md)
</details>


aaltshuler:
Rebased this PR onto current `main` and pushed the initialization-safety fixes to the contributor branch.

The updated implementation now:

- acquires one durable root-scoped init claim for strict and `--force` callers, so concurrent initializers cannot overwrite or clean up one another;
- treats every error after physical dataset initialization begins as acknowledgement-unknown and preserves the schema contract;
- authenticates an exact attempt-local genesis receipt after a lost manifest acknowledgement;
- distinguishes committed, indeterminate, and already-claimed initialization outcomes with typed errors;
- refuses strict or forced init when an existing `__manifest` is present;
- retains the claim when cleanup or physical outcome is indeterminate;
- documents the transient `__init_claim.json` object and the narrowed `--force` behavior.

The new tests cover force/force and force/strict races, table- and manifest-Create lost acknowledgements, exact-genesis mismatch, failed outcome probes, cleanup-delete ambiguity, stale claims, and fresh reopen after recovered acknowledgement.

Validation on the final tree:

- canonical workspace failpoint-superset: passed (with the repository-required macOS debug stack setting);
- default and failpoint-superset all-target clippy: passed;
- rustfmt, AGENTS link check, focused init/failpoint/lifecycle/forbidden-API suites, CLI parity, server/OpenAPI suites: passed;
- fresh GitHub required CI: passed;
- fresh Greptile review: 18 files reviewed, 0 comments added.

PR #514, opened 2026-08-15
fix: retry raced Lance branch enumeration

### LINKED ISSUE #515: bug: operations racing branch_delete fail with an unclassified "Not found" branch-ref error
### What happened

A `branch_delete` running concurrently with almost any other write-path operation on the same store can make the OTHER operation fail with a raw, unclassified internal error naming a ref file of the branch being deleted. Both operations are legal public API calls; no fault injection or crash is involved. Expected: either a typed retryable error, or the operation tolerating a branch that vanished mid-enumeration.

Mechanism, verified in source at main `db23f58`. Many write-path operations enumerate branches on the `__manifest` control dataset through Lance's `Dataset::list_branches()`. Inside Lance, `list_branches` first lists the `_refs/branches/` directory and then reads each ref file it found (`from_path`, `lance/src/dataset/refs.rs:915-928`; identical code in Lance 9.0.0 and 10.0.0). `branch_delete` physically removes that ref file. A delete landing between the internal listing and the ref read makes the read fail with `Not found: .../_refs/branches/NAME.json`. This is a time-of-check-to-time-of-use race entirely inside one `list_branches` call, so no caller-side check can close it. The engine then maps every `list_branches` error into the catch-all string form `OmniError::Lance(e.to_string())` (`crates/omnigraph/src/db/manifest.rs:1159`, `:1192`, `:1205`), so the caller receives an unclassified internal error for a routine race between two legal operations. A client cannot distinguish "a branch was deleted, retry" from "store is broken".

Three distinct victim operations were observed (measured, concurrent test runs): a plain data mutation on main (`refs.rs:920` in the error), `optimize` (`refs.rs:923`), and `ensure_indices` (`refs.rs:923`). The enumeration is shared write-path machinery, not specific to one operation. Any deployment where an operations cron runs maintenance while agents create and delete branches walks this window.

The engine already contains both halves of a fix, applied today only to `cleanup`: cleanup acquires the branch guards for ALL branches before its branch work (`crates/omnigraph/src/db/omnigraph/optimize.rs:1087`) and wraps its per-branch classification reads in a typed refusal with a remedy (`optimize.rs:1126`). Other enumerating operations gate main only (for example `optimize` takes the schema gate, the main branch guard, and main table queues, `optimize.rs:246-263`) yet still enumerate all branches. Fix directions, cheapest first: classify a vanished-ref failure from branch enumeration as a typed retryable conflict (the `optimize.rs:1126` precedent); or make enumeration tolerate a ref vanishing between list and read, since a deleted branch needs no work (in Lance this is one error arm: `Branches::fetch` streams the per-ref reads and fails the whole listing on the first failed read, `lance/src/dataset/refs.rs:395-408` in 10.0.0 — skipping a Not-found ref there fixes every caller, including Lance's own cleanup scans that never pass through this engine's error mapping); or extend branch-guard acquisition to enumerating operations, which closes only the in-process case because cross-process writers share no gates.

### Steps to reproduce

The window is a few storage operations wide inside one `list_branches` call. We reproduced it only under heavy parallel load, never in isolation; details and non-reproducing shapes below so triage does not chase them.

1. Build from source and use the in-memory storage backend (any backend should work; the window is list-then-read against the object store).
2. Run 12 concurrent test processes. In each process, run universes of: 3 writer threads doing single-row upsert mutations on main (about 30 ops each), plus one branch-actor thread cycling `branch_create` -> 3 inserts on the branch -> `branch_merge` to main -> `branch_delete`, 3 cycles; optionally one maintenance thread alternating `optimize` / `cleanup` / `ensure_indices`.
3. Observed (measured): 6 failing universes across roughly 7,000 universes in four such runs, i.e. well under 1 percent of branch-bearing universes. Each failure is one of the error lines in Logs, surfaced to the caller of the victim operation.
4. Shapes that did NOT reproduce (measured negatives, so triage can skip them): single-process two-thread enumerate-vs-delete loops (80 rounds, empty and data-bearing branches); the same universe shape at 30 seeds inside one process, with both single-threaded and 4-thread Lance pools. In single-process runs the engine's write gates were observed to mutually exclude `branch_delete`'s delete phase and other writers' storage activity, so the window appears to need the thread-pool parallelism and preemption that only machine-wide contention produced in our runs.
5. The reproducing build carried deterministic-testing seams (injected id and clock sources) that do not alter locking or refs traffic.

### Version

omnigraph 0.10.0, built from source, main @ e53f3c9 (Lance 9.0.0); window re-verified in source at main @ db23f58 (Lance 10.0.0, refs.rs unchanged)

### Environment

macOS (Darwin 25.5) arm64; in-memory storage adapter (registered shared-memory scheme). Expected backend-independent: branch refs are JSON objects in the store and the list-then-read window exists against any object store.

### Logs / output

```shell
Lance("Not found: __manifest/_refs/branches/cb1.json, .../lance-9.0.0/src/dataset/refs.rs:920:18")
Lance("Not found: __manifest/_refs/branches/cb2.json, .../lance-9.0.0/src/dataset/refs.rs:923:18")
```

azimafroozeh:
## What & why

Lance enumerates named branches by listing `_refs/branches/` and then reading every listed ref. If a branch is deleted between those reads, the otherwise read-only enumeration can fail with `NotFound`.

This change retries only that read-only enumeration, with a fixed limit of three total attempts. It does not turn a storage error into permission to replay the enclosing operation, and it does not introduce a new HTTP or OpenAPI error contract. Every production `Dataset::list_branches()` call now goes through the same helper.

## Backing issue / RFC

- [x] Fixes an **accepted** issue: Closes #515
- [ ] Is an RFC PR, or implements a **merged** RFC
- [ ] **Trivial fast-lane**

The retry boundary is aligned with RFC-038: the engine retries the known effect-free read locally instead of exposing a generic whole-operation retry signal.

## Checklist

- [x] Tests added/updated for behavior changes
- [x] Public docs updated if user-facing surface changed, or N/A — no public surface change
- [x] Reviewed against `docs/dev/invariants.md`

## Local verification

- `cargo fmt --all --check` — passed
- `scripts/check-agents-md.sh` — passed
- `cargo test -p omnigraph-engine --locked --lib branch_control::tests -- --nocapture` — 10 passed
- `cargo test -p omnigraph-engine --locked --test forbidden_apis` — 18 passed
- `cargo clippy -p omnigraph-engine --all-targets --locked -- -D warnings -W clippy::dbg_macro` — passed

## Notes for reviewers

The regression test uses Lance's `file-object-store://` provider and an object-store wrapper to physically delete a ref after it has been listed but before Lance reads it. It proves that the next attempt performs a fresh listing and succeeds. A second test holds the ref in a repeated `NotFound` state and proves the exact three-attempt bound and final substrate error.

There is intentionally no 409 mapping or instruction for callers to retry an effectful operation. Repeated churn escapes through the ordinary storage-error path for the owning operation to classify.

<h3>Greptile Summary</h3>

The PR centralizes Lance branch enumeration behind a bounded, read-only retry that handles refs deleted between listing and reading.

- Retries `NotFound` enumeration failures up to three total attempts.
- Routes manifest, recovery, indexing, and native table-branch enumeration through the helper.
- Adds race and exhaustion regression tests plus a structural guard preventing raw production calls outside the helper.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph/src/branch_control.rs | Adds the bounded branch-enumeration retry boundary and focused tests for a vanished ref and exact retry exhaustion. |
| crates/omnigraph/src/db/manifest.rs | Routes manifest branch listing, descendant discovery, and delete preflight through the centralized helper. |
| crates/omnigraph/src/db/manifest/recovery.rs | Routes recovery-time branch observations through the retry helper while preserving existing gates and identity checks. |
| crates/omnigraph/src/storage_layer.rs | Renames the sealed storage method to distinguish native Lance branches from graph-level branches. |
| crates/omnigraph/src/table_store.rs | Implements native branch enumeration through the centralized retry helper. |
| crates/omnigraph/tests/forbidden_apis.rs | Updates renamed gateway inventories and structurally enforces a single raw `Dataset::list_branches` production call site. |

</details>

```mermaid
sequenceDiagram
    participant Caller
    participant Helper as list_branch_contents
    participant Lance
    participant Store as Object store
    Caller->>Helper: enumerate branches
    loop At most three attempts
        Helper->>Lance: Dataset::list_branches()
        Lance->>Store: list _refs/branches/
        Store-->>Lance: listed refs
        Lance->>Store: read each ref
        alt Ref disappears after listing
            Store-->>Lance: NotFound
            Lance-->>Helper: NotFound
            Helper->>Helper: yield and retry
        else Enumeration succeeds
            Store-->>Lance: branch contents
            Lance-->>Helper: branch map
            Helper-->>Caller: branch map
        else Other error or final NotFound
            Lance-->>Helper: substrate error
            Helper-->>Caller: OmniError::Lance
        end
    end
```
</details>

<sub>Reviews (3): Last reviewed commit: ["test(branch): guard enumeration retry bo..."](https://github.com/modernrelay/omnigraph/commit/1ffb2de6c53906c14b7f0eaa215dfed91eb6ae8a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53663478)</sub>

**Context used:**

- Knowledge Base — [Branching and merge](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/branching-and-merge.md)


azimafroozeh:
Closes #515.

PR #522, opened 2026-08-16
perf(changes): bound candidate scans to transaction footprints

ragnorc:
## Summary

- prune only exact adjacent Lance versions whose pinned child transaction reads the published parent version
- derive new child fragments and touched parent fragments from the transaction using binary search, without whole-manifest set walks
- stream child candidates and parent before-images in id order with one prepared row per side
- remove BTREE correctness and cost dependence, so absent and stale id indexes behave identically
- size scans from the remaining page row/byte budget and stop on the sentinel before image, JSON, or Blob materialization
- fall back conservatively for multi-version intervals, missing proof metadata, unsupported transactions, deletes, or any structural mismatch

## Review fixes

This revision addresses the reported DX/performance cliffs:

- no 8,192-row candidate queue on max_changes=1
- no full candidate-batch VecDeque
- no full parent/child fragment HashSet diff
- no per-page transaction-history walk for non-adjacent intervals
- no point-lookup cliff when the scalar index is absent or partially covered

New counters pin transaction reads, fragment metadata comparisons, candidate rows examined, scan targets, and image materialization.

## Verification

- cargo fmt --all --check
- cargo clippy --workspace --all-targets --locked -- -D warnings -W clippy::dbg_macro
- cargo clippy --workspace --all-targets --locked --features omnigraph-engine/failpoints,omnigraph-cluster/failpoints -- -D warnings -W clippy::dbg_macro
- cargo test -p omnigraph-engine --test changes --test changes_cost --locked
- cargo test -p omnigraph-engine --features failpoints --test failpoints change_feed_poll_classifies_intervals_before_the_head_witness --locked -- --exact
- cargo test -p omnigraph-engine --test forbidden_apis --locked
- cargo test --workspace --locked --features omnigraph-engine/failpoints,omnigraph-cluster/failpoints
- scripts/check-agents-md.sh

<h3>Greptile Summary</h3>

The PR optimizes adjacent commit-change enumeration by deriving candidate scans from the published Lance transaction footprint while retaining the full ordered merge as a conservative fallback.

- Adds transaction-provenance, fragment-shape, and row-version metadata gates for the pruned path.
- Streams candidate and parent rows in ID order with page-sized scan targets and lazy row preparation.
- Adds instrumentation, failpoint coverage, cost assertions, and documentation for the new planning path.

The PR appears safe to merge because no blocking failure remains in the eligible follow-up review scope.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph/src/changes/candidate_scan.rs | Introduces the guarded transaction-footprint planner and streaming candidate emitter, with conservative fallback for unsupported or incomplete physical evidence. |
| crates/omnigraph/src/changes/enumerate.rs | Integrates pre-witness candidate planning and sentinel-aware page budgeting without an accepted new correctness finding. |
| crates/omnigraph/src/changes/row_compare.rs | Replaces eager per-batch row preparation with a lazy cursor and adds fragment-scoped scans with page-derived targets. |
| crates/omnigraph/src/table_store.rs | Adds durable transaction provenance used to distinguish engine-authored row-set-preserving updates from untrusted update shapes. |
| crates/omnigraph/tests/changes_cost.rs | Expands cost coverage for bounded transaction reads, fragment comparisons, candidate scans, and image materialization. |

</details>

```mermaid
flowchart TD
  A[Changed table interval] --> B{Adjacent versions and same lineage?}
  B -- No --> F[Full ordered merge]
  B -- Yes --> C[Read pinned child transaction]
  C --> D{Proven row-set preserving and structurally valid?}
  D -- No --> F
  D -- Yes --> E[Build touched child and parent fragment plan]
  E --> G[Stream candidates and before-images by ID]
  F --> H[Emit exact logical changes]
  G --> H
  H --> I[Apply page budget and continuation]
```
</details>

<sub>Reviews (10): Last reviewed commit: ["chore(ci): classify change-scan vocabula..."](https://github.com/modernrelay/omnigraph/commit/628fa400da3892481aad1f31685ef457248f54ee) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53795976)</sub>

- Knowledge Base — [Branching and version history](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/branching-and-version-history.md)
- Knowledge Base — [Data mutation pipeline](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/data-mutation-pipeline.md)
</details>


ragnorc:
Rebased onto the updated `change-feed`.

The base branch replaced the `rows_equal` scope hint (`BlobComparisonScope`) with an unconditional identity qualifier: managed-Blob descriptors are now keyed by the owning data file's immutable path (a globally-unique per-file UUID), which is authoritative across lineage without a same-branch/cross-branch distinction. The candidate-scan parent probe therefore calls the scope-free `rows_equal` directly, and the interim "adapt … scope" commit (which existed only to pass `SameLineage` to the old signature) is dropped as obsolete. `build_change_feed_cut` is already registered in the surface guard on the base.

No behavior change to the pruning logic. Verified against the rebased base: engine builds, and `changes` (42), `changes_cost` (6, including the pruned-FLAT / fallback-GROWING gates), and `forbidden_apis` (18) all pass.

crates/omnigraph/src/changes/candidate_scan.rs
```diff
@@ -0,0 +1,408 @@
...
+pub(crate) fn operation_is_row_set_preserving(operation: &Operation) -> bool {
+    match operation {
+        // Append only adds fragments; it never removes or reuses a logical id.
+        Operation::Append { .. } => true,
+        // OmniGraph's keyed writes (strict-insert / upsert / known-present
+        // update) are all `RewriteRows` merge_insert and never delete an
+        // unmatched-by-source row — no delete-capable by-source merge arm exists
+        // in the engine (locked by the write-path guard in forbidden_apis.rs). A
+        // different update mode is a foreign or unknown shape, so fall back.
+        Operation::Update { update_mode, .. } => update_mode == &Some(UpdateMode::RewriteRows),
```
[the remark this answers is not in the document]
ragnorc: Confirmed — this is a real gap. The candidate path's changed-fragment scan has no delete pass, and `repair --force --confirm` can adopt an external Lance merge whose delete-capable by-source arm persists as `Operation::Update { RewriteRows }`. The `forbidden_apis` source guard only proves that *current engine code* builds no such arm — it cannot authenticate persisted transaction history.

Two candidate fixes, and why one is insufficient:

- **Gate on the RFC-023 `insert_absence` certificate** — insufficient here. #522's core pruned case (`changes_page_opens_and_data_reads_are_bounded_by_delta`) is an **upsert** that updates an existing row; it carries no `insert_absence` certificate, so this gate would fall the whole optimization back for the exact case it targets. (Row-count preservation is also insufficient: a delete+insert nets zero.)
- **Durable OmniGraph keyed-write provenance marker** — the correct fix. Stamp a `no-by-source-delete` property on every keyed-write `Update` transaction (the keyed writer structurally has no delete-capable by-source arm) and require it for `Update` pruning; external merges adopted via `repair --force` lack the marker and fall back to the exact merge. The marker is read-advisory (a missing one only costs the optimization, never correctness), so it composes with recovery.

This is a required pre-merge change on this PR; it needs a small write-path addition plus a classifier gate and a regression test that a marker-less `Update{RewriteRows}` in the interval forces the exact-merge fallback. Leaving open.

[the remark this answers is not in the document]
ragnorc: Fixed in 521240bd / 20f8017a (docs d9288dfb).

The candidate-pruning classifier no longer trusts an `Update{RewriteRows}` on its op shape alone. The write path now stamps a durable `omnigraph.no_by_source_delete` transaction property at the single general keyed-merge chokepoint (`staged_keyed_merge_result`), so every OmniGraph keyed `Update` (upsert, known-present update, stream strict insert) carries a per-transaction proof that it removed no unmatched-by-source row. The new `transaction_is_row_set_preserving` gate requires that marker — or the RFC-023 `insert_absence` certificate — before pruning an `Update`; `Append` stays unconditional. A marker-less `Update` (exactly what an external delete-capable merge adopted via `repair --force` persists) falls back to the exact ordered merge, whose delete pass reports the removed rows the child-only candidate scan would miss.

Notes:
- The marker is read-advisory: stamped unconditionally, it survives commit and recovery (it lives in Lance's committed manifest; recovery reuses the landed version), and a missing marker only forces the fallback, never a correctness change.
- It is distinct from `insert_absence` (minted only for pure inserts), which is why gating on `insert_absence` alone was rejected — #522's core pruned case is a real update-bearing upsert that carries the marker but not the certificate.
- The `no_delete_capable_merge_arm_in_engine_source` source guard is retained as defense-in-depth; the marker is the stronger per-transaction proof it could never provide.

Tests: `table_store::staged_tests::keyed_upsert_stamps_no_by_source_delete_marker_and_persists_it` (write path stamps + survives commit → `list_transactions`), `candidate_scan::transaction_update_prunes_only_with_a_durable_no_delete_proof` (marker-less `Update` is not row-set-preserving), and the existing upsert-prune cost/image tests stay green as live end-to-end proof the stamp and gate work together. Full failpoint-superset workspace suite green.

crates/omnigraph/src/changes/candidate_scan.rs
```diff
@@ -0,0 +1,558 @@
...
+    // error.
+    let Ok(delta) = to_dataset
+        .delta()
+        .with_begin_version(from_entry.table_version)
+        .with_end_version(to_entry.table_version)
+        .build()
+    else {
+        return Ok(None);
+    };
+    let Ok(transactions) = delta.list_transactions().await else {
```
aaltshuler: [P1] Re-prove the named-branch incarnation after this live history read. The existing final witness runs before `EmitSource::plan`, while Lance `DatasetDelta::list_transactions()` re-checks out every numeric version through the current `<root>/tree/<branch>` path. A delete/recreate can therefore supply replacement safe-looking transactions while `changed` is computed from the old pinned manifests; old whole-fragment Delete plus replacement Append/marked Update selects an empty candidate scan and silently advances past the delete. Materialize candidate eligibility before the existing final reproof, or repeat the witness after all transaction-history reads, and add an ABA rendezvous regression that defeats Lance's incidental transaction cache.

crates/omnigraph/src/changes/candidate_scan.rs
```diff
@@ -0,0 +1,558 @@
...
+
+/// Whether one changed fragment's `_row_last_updated_at_version` sequence is
+/// present and decodable — the §4.2 "genuinely active" requirement at the
+/// fragment level. `uses_stable_row_ids()` alone is a dataset-level flag and
+/// cannot prove a specific fragment's sequence survives loading.
+fn fragment_version_metadata_is_loadable(fragment: &Fragment) -> bool {
+    fragment
+        .last_updated_at_version_meta
+        .as_ref()
+        .is_some_and(|meta| meta.load_sequence().is_ok())
```
aaltshuler: [P1] This does not safely prove that version metadata is loadable and structurally valid. In pinned Lance 10, `RowDatasetVersionMeta::External` is a real persisted variant but `load_sequence()` executes `todo!()`, so this read path panics instead of falling back. For Inline metadata, successful protobuf decoding also does not prove `sequence.len() == fragment.physical_rows`; Lance's single-run fast path ignores the encoded run length and stamps the run version over all requested rows, which can silently move rows into or out of the candidate window. Match Inline explicitly, require an exact physical-row length, and return false for External/None/error/mismatch.

crates/omnigraph/src/changes/candidate_scan.rs
```diff
@@ -0,0 +1,558 @@
...
+    }
+
+    async fn next(&mut self) -> Result<Option<Emit>> {
+        loop {
+            if let Some(emit) = self.ready.pop_front() {
+                return Ok(Some(emit));
+            }
+            // Pull the next id-ordered chunk of candidates, then probe the
+            // parent once for the whole chunk.
+            let mut chunk: Vec<RawRow> = Vec::new();
```
aaltshuler: [P1] Bound this queue by retained bytes and the current page budget, not only 8,192 rows. Before yielding one change, this holds all child `RawRow`s, cloned IDs, the parent HashMap, and every queued Emit; each one-row Arrow slice can retain its full scanner batch. `max_changes=1` still probes and compares all 8,192 candidates (including Blob payload tie-breaks), then truncation drops the remaining queue and the next page repeats the work. Large/multi-version eligible intervals can therefore exceed the page memory contract and amplify continuation I/O dramatically. Please add large-delta, page-size-1 byte/RSS and Blob-work coverage.

crates/omnigraph/src/changes/candidate_scan.rs
```diff
@@ -0,0 +1,558 @@
...
+        .as_ref()
+        .is_some_and(|meta| meta.load_sequence().is_ok())
+}
+
+/// Fetch full before-image rows for `ids` from the parent handle in one
+/// BTREE-backed `id IN (chunk)` lookup (never per-row round trips, never a
+/// string filter). The returned rows carry `_rowid`/`_rowaddr` and Blob
+/// descriptions so `rows_equal` and `emitted_image` behave exactly as on the
+/// full-merge path.
+async fn probe_parent_images(parent: &Dataset, ids: &[String]) -> Result<HashMap<String, RawRow>> {
```
aaltshuler: [P2] This is only BTREE-backed when the parent index fully covers every current fragment. Missing and stale derived indexes are normal; in that state Lance scans uncovered/all parent fragments once per chunk, so this path is still O(parent extent) and can become slower than the one-pass fallback. Gate the strategy on the existing `TableStore::key_column_index_coverage(..., "id") == IndexCoverage::Indexed` check (or use a one-pass alternative), and add absent/partial-index cost cells. The current flat test hides the cliff by calling `ensure_indices()` immediately before measurement.

crates/omnigraph/src/changes/candidate_scan.rs
```diff
@@ -0,0 +1,558 @@
...
+        return Ok(None);
+    }
+
+    // The changed fragments are exactly the child fragments the parent does not
+    // have: every proven op only ADDS fragments (append, or rewrite = remove old
+    // + add new), so a child fragment absent from the parent carries this
+    // interval's inserted/updated rows. Computed from the already-loaded
+    // manifests — no object-store data reads — so the candidate scan reads only
+    // O(delta) fragments.
+    let parent_ids: std::collections::HashSet<u64> = from_dataset
```
aaltshuler: [P2] The implementation still walks and allocates over every parent fragment and then every child fragment, so CPU and retained metadata are O(dataset fragment count), not O(delta). The new cost test counts data-read operations only and cannot detect this term. Derive the surviving changed-fragment set by folding the already-loaded transaction operations, or narrow the advertised asymptotic and add a fragment-count/CPU-memory curve.

crates/omnigraph/src/changes/candidate_scan.rs
aaltshuler: Verified fixed at `4eaa4cf`. The live `(begin, end]` transaction-history walk now completes inside `plan_intervals` before both final `reprove_named_branch_heads` witnesses, and `EmitSource::plan` consumes the saved fragment decision without reopening numeric-path history. The new post-witness failpoint recreates the branch at the same versions and accepts only the original delete or a loud refusal. This closes the ABA window; resolving.

crates/omnigraph/src/changes/candidate_scan.rs
aaltshuler: Verified fixed at `00e2d2ea`. The gate now structurally accepts only `RowDatasetVersionMeta::Inline`, decodes it directly, and requires `sequence.len() == physical_rows`; External, absent, undecodable, missing-count, short, and long sequences all fall back to the exact merge. The external and short/exact/missing-count regressions pin the boundary. Resolving.

crates/omnigraph/src/changes/enumerate.rs
```diff
@@ -329,12 +341,33 @@ async fn plan_intervals(
...
                 let (kind, type_name) = parse_table_key(table_key);
+                let kind: ChangeEntityKind = kind.into();
+                // Classify the interval HERE — before the head witness below —
+                // because the classifier's `(begin, end]` transaction walk is a
+                // live read of replaceable numeric-path version manifests.
+                // Scope-filtered intervals are never emitted, so they skip the
+                // walk; their stored decision is irrelevant.
+                let changed_fragments =
+                    if scope.wants_kind(kind) && scope.wants_type_name(type_name) {
+                        super::candidate_scan::interval_changed_fragments(
```
aaltshuler: [P2] This correctness move places the history read under the final witness, but it still runs once per stateless page. Every continuation reconstructs `plan_intervals` and calls `interval_changed_fragments`, which re-lists/checks out the complete `(begin, end]` history (up to 1,024 versions) before returning even one change. With `max_changes=1`, a multi-version interval therefore pays the full history cost on every page, contrary to the advertised bounded O(delta) pagination. Please make this proof reusable through a safe immutable/pinned plan (not a forgeable trust bit in the checksum-only token), or narrow the claim and add a multi-version, page-size-1 transaction-read curve.

aaltshuler (commented):
Follow-up review of `4eaa4cf`: the row-version metadata and post-witness branch-ABA P1s are correctly fixed and their threads are resolved. The PR is still not merge-ready: the 8,192-candidate queue remains unbounded by the page row/byte budget; full-index coverage is still assumed; fragment discovery remains O(total fragments); and the inline comment below covers repeated transaction-history work on continuation. CI is currently red on two trivial `needless_borrow` Clippy errors. The PR also still targets the already-merged `change-feed` branch; it must be rebuilt/retargeted onto current `main` and revalidated there.

crates/omnigraph/src/changes/candidate_scan.rs
```diff
@@ -0,0 +1,558 @@
...
+                            before: before.clone(),
+                            after: candidate,
+                        }
+                    }
+                };
+                if self.scope.wants_op(emit.op()) {
+                    self.ready.push_back(emit);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
aaltshuler: Fixed in 385055bc. This is the same queue/page-budget cliff now removed from CandidateUpserts and OrderedRows: no ready VecDeque, no parent HashMap, and at most one prepared RawRow per stream. Candidate scan targets are remaining_rows + one sentinel and remaining page bytes. The 2,048-row max_changes=1 regression proves two examined candidates and one image materialization on each of two stateless pages; the canonical failpoint-superset workspace suite is green.

crates/omnigraph/src/changes/candidate_scan.rs
aaltshuler: Fixed in 385055bc without making the optimization depend on derived index state. Parent point probes are gone. CandidatePlan now carries only parent fragments named by the pinned Update transaction, OrderedRows scans those fragments once in id order, and CandidateUpserts lockstep-merges that stream with candidate child rows. changes_page_opens_and_data_reads_are_bounded_by_delta now runs asserted absent-index and stale/partial-index cells; both remain Degraded and both have the same flat data-read cost.

crates/omnigraph/src/changes/candidate_scan.rs
aaltshuler: Fixed in 385055bc. The full parent/child HashSet walk is gone. The child suffix is located by binary search above the parent manifest high-water mark and validated against the transaction new-fragment count/sequence; only removed_fragment_ids plus updated_fragments are deduplicated, and each touched parent fragment is found by binary search. fragment_discovery_is_binary_search_plus_delta uses 65,536 fragments and pins each lookup at at most 17 comparisons. The RFC claim is narrowed accordingly to metadata O(log F + T log F), with data work scoped to new plus transaction-touched parent fragments.

crates/omnigraph/src/changes/candidate_scan.rs
aaltshuler: Fixed in 385055bc. CandidateUpserts no longer builds an 8,192-row probe queue or parent HashMap: it merges one lazily prepared candidate row with one lazily prepared touched-parent row. ScanTargets::for_page derives both row and byte targets from the remaining page budget plus one continuation sentinel, and enumerate stops on that sentinel before image/JSON/Blob materialization. changes_page_size_one_bounds_large_candidate_delta pins two stateless max_changes=1 pages over a 2,048-Blob-row delta at two examined rows, a row target of two, and one materialized image per page; the existing Blob-work test also remains green.

crates/omnigraph/src/changes/enumerate.rs
aaltshuler: Fixed in 385055bc by making the optimization adjacent-only. interval_candidate_plan rejects any interval where child_version != parent_version + 1 before reading a transaction, so multi-version intervals use the exact ordered merge and never list/check out transaction history. changes_page_size_one_skips_transaction_history_for_multi_version_intervals publishes two Lance updates under one graph interval, resumes across two stateless size-one pages, and asserts candidate_transaction_reads == 0 on both pages. The former broad O(delta) title/docs claim has also been narrowed.

PR #531, opened 2026-08-18
perf(mutate): stage independent tables at the loader's write concurrency

### LINKED ISSUE #504: mutate stages tables serially: measured ~0.47s/table on S3, and stage_all_with_concurrency already exists
## Summary

`mutate` stages tables serially, so a single multi-table mutation pays one sequential round of object-store PUTs per table. `stage_all_with_concurrency` already exists and the loader already uses it — `stage_all` just pins concurrency to `1`.

`crates/omnigraph/src/exec/staging.rs` (v0.9.0):

```rust
/// Sequential per-table for now — parallelizing across independent
/// Lance datasets is a perf follow-up; same loop structure as the
/// pre-split `finalize`.
pub(crate) async fn stage_all(
    self,
    db: &crate::db::Omnigraph,
    branch: Option<&str>,
) -> Result<StagedMutation> {
    self.stage_all_with_concurrency(db, branch, 1).await
}
```

Filing because we measured what the follow-up is worth against S3, in case that helps prioritise it.

## Measurement

Deployed omnigraph-server v0.9.0, Lance-on-S3, single graph. The client writes one node plus its tag nodes and edges as **one** `mutate` — so the commit count is already 1 and the only variable is how many tables that commit touches.

| write | tables touched | median |
|---|---|---|
| node only | 1 | **2.79s** |
| node + tags + edges | 3 | **3.73s** |

Alternating variants, 3 samples each, same graph and session. The delta is **+0.94s for two additional tables ≈ 0.47s/table**, which matches one sequential PUT round-trip per table.

So on this deployment ~25% of a 3-table mutation is serial staging that has no data dependency between the tables.

## Why it matters to us

Our writes sit behind a proxy with a hardcoded 30s client deadline. Under 16 concurrent writers to one graph the per-write cost stretches to 17.5s median / 33.6s tail, at which point the proxy cuts the connection and returns 502 **while the write goes on to commit successfully**. The caller is told it failed, the row is there, and nothing in the reply distinguishes that from a genuine failure.

Serial staging is not the dominant term in that — contention on the commit cycle is — but it is a consistent ~25% on every multi-table write, and it is the part with an existing mechanism to fix it.

## Suggestion

Have `stage_all` take the same concurrency the loader does (`OMNIGRAPH_LOAD_CONCURRENCY`, or a separate knob if mutation should be tuned independently), defaulting to something >1. The independent-Lance-dataset property the comment relies on is the same one the loader path already assumes.

Happy to test a branch against our workload — we have a reproducible concurrency probe and a populated graph, and can report before/after numbers on the same deployment.

## Environment

- omnigraph-server v0.9.0, storage format 6, Lance on S3
- Kubernetes, single graph, writes arriving via an MCP tier

memmmmike:
Closes #504.

Per the acceptance comment: the fix stays inside the existing mechanism. `stage_all` now delegates at `stage_write_concurrency()` (the loader's `OMNIGRAPH_LOAD_CONCURRENCY` knob, default 8) instead of a pinned 1. The resolver moves to `exec::staging` with a pure parse half; the loader's private copy is deleted and both call sites share it.

**Scope kept, verified against source:**
- Publication untouched — everything after staging still funnels through the single manifest CAS; `commit_all` still acquires sorted table gates.
- Ordering untouched — the stream is `buffered` (order-preserving), not `buffer_unordered`.
- Failure semantics untouched — the staging stream drained before surfacing the first error at width 1 too (`buffered(1).collect::<Vec<Result<_>>>`), so wider staging changes only how much of that pre-existing drain overlaps; residue remains reclaimable and never graph-visible.

**Tests:** a pure unit test on the parse rules, and a serial-vs-concurrent equivalence test on a multi-table mutation (Person + two Knows edges in one query) asserting affected counts, per-table row counts, the inserted row's values, and exact edge endpoint pairs agree between `OMNIGRAPH_LOAD_CONCURRENCY=1` and `=4` (`#[serial]`-guarded per the search.rs precedent). Full `writes` suite: 43/43.

**Ad-hoc local measurement** (16 node tables touched per mutation, MinIO behind a 25ms toxiproxy latency, 5 timed runs each after a warm-up): end-to-end `mutate` mean 5.80s at concurrency 1 vs 4.00s at 8. The remaining gap to the theoretical staging ratio is the untouched validate/publish path, which stays serial by design. Happy to re-run under any harness you prefer, per docs/dev/testing.md's benchmark guidance.

https://claude.ai/code/session_01E4S9Xpe9G7pLEv9JaoXUAb

<h3>Greptile Summary</h3>

The PR makes constructive multi-table mutation staging use the loader’s bounded write concurrency while preserving serial publication and recovery boundaries.
- Centralizes staging-concurrency resolution for loader and mutation paths.
- Adds task-local test controls and deterministic overlap probes without mutating process-wide environment.
- Documents the expanded configuration scope and adds concurrency-equivalence and write-cost coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph/src/exec/staging.rs | Shares the loader concurrency resolver with ordinary constructive mutations while leaving deferred forks, deletes, commit ordering, and publication unchanged. |
| crates/omnigraph/src/instrumentation.rs | Adds task-local concurrency overrides and deterministic staging probes for isolated concurrency testing. |
| crates/omnigraph/src/loader/mod.rs | Replaces the loader-private concurrency resolver with the shared staging resolver without changing its default or parsing behavior. |
| crates/omnigraph/tests/writes.rs | Replaces the process-wide environment mutation with scoped controls and verifies serial-versus-concurrent result equivalence and actual overlap. |
| crates/omnigraph/tests/write_cost.rs | Adds coverage ensuring multi-table staging read costs remain flat as commit history grows. |
| docs/dev/writes.md | Documents the shared staging width while preserving the single-publication and recovery model. |
| docs/user/mutations/index.md | Documents the environment variable’s expanded mutation scope, default, and exclusions. |
| docs/releases/v0.10.0.md | Records bounded multi-dataset mutation staging as a user-visible release change. |

</details>

```mermaid
flowchart LR
  A[Validated mutation or load] --> B[Stage independent existing tables concurrently]
  B --> C[Acquire ordered table gates]
  C --> D[Record recovery intent]
  D --> E[Commit table effects]
  E --> F[Publish one graph manifest CAS]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(mutate): prove bounded concurrent ta..."](https://github.com/modernrelay/omnigraph/commit/02b50092d9f8b535a63c70c97cd53645134f463a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54194312)</sub>

- Knowledge Base — [Data mutation pipeline](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/data-mutation-pipeline.md)
- Knowledge Base — [Recovery and durability](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/recovery-and-durability.md)
</details>

### DIFF
--- a/crates/omnigraph/src/exec/staging.rs
+++ b/crates/omnigraph/src/exec/staging.rs
@@ -168,7 +168,16 @@ pub(crate) struct MutationStaging {
 /// Override via `OMNIGRAPH_LOAD_CONCURRENCY`.
 pub(crate) const DEFAULT_STAGE_WRITE_CONCURRENCY: usize = 8;
 
+/// Resolution order: the scoped test override
+/// ([`crate::instrumentation::with_stage_write_concurrency`]), then
+/// `OMNIGRAPH_LOAD_CONCURRENCY`, then the default. Tests force a width through
+/// the scoped seam so they never mutate process-global environment.
 pub(crate) fn stage_write_concurrency() -> usize {
+    if let Some(scoped) = crate::instrumentation::stage_write_concurrency_override()
+        && scoped > 0
+    {
+        return scoped;
+    }
     parse_stage_write_concurrency(std::env::var("OMNIGRAPH_LOAD_CONCURRENCY").ok().as_deref())
 }
 
@@ -1671,14 +1680,51 @@ fn dedupe_merge_batches_by_id(
 
 #[cfg(test)]
 mod stage_write_concurrency_tests {
-    use super::parse_stage_write_concurrency;
+    use super::{parse_stage_write_concurrency, stage_write_concurrency};
+    use crate::instrumentation::with_stage_write_concurrency;
 
     // Pure parser, no process-global environment touched.
     #[test]
     fn resolves_default_override_and_junk() {
-        assert_eq!(parse_stage_write_concurrency(None), 8, "default without the env var");
+        assert_eq!(
+            parse_stage_write_concurrency(None),
+            8,
+            "default without the env var"
+        );
         assert_eq!(parse_stage_write_concurrency(Some("3")), 3, "override wins");
-        assert_eq!(parse_stage_write_concurrency(Some("0")), 8, "zero is not a concurrency");
-        assert_eq!(parse_stage_write_concurrency(Some("banana")), 8, "junk falls back to default");
+        assert_eq!(
+            parse_stage_write_concurrency(Some("0")),
+            8,
+            "zero is not a concurrency"
+        );
+        assert_eq!(
+            parse_stage_write_concurrency(Some("banana")),
+            8,
+            "junk falls back to default"
+        );
+    }
+
+    /// The scoped seam must actually reach the resolver, and must not outlive its
+    /// future. Without this, a width-forcing test would silently run at the default
+    /// width on both sides of an equivalence comparison and prove nothing.
+    #[tokio::test]
+    async fn scoped_override_wins_and_does_not_leak() {
+        let outside_before = stage_write_concurrency();
+        assert_eq!(
+            with_stage_write_concurrency(3, async { stage_write_concurrency() }).await,
+            3,
+            "the scoped override must reach the resolver"
+        );
+        // 0 is not a concurrency: the scope is ignored, not obeyed.
+        assert_ne!(
+            with_stage_write_concurrency(0, async { stage_write_concurrency() }).await,
+            0,
+            "a zero override must fall back, never pin the width to zero"
+        );
+        assert_eq!(
+            stage_write_concurrency(),
+            outside_before,
+            "the override must be gone once its future resolves"
+        );
     }
 }
--- a/crates/omnigraph/src/exec/staging.rs
+++ b/crates/omnigraph/src/exec/staging.rs
@@ -163,9 +163,8 @@ pub(crate) struct MutationStaging {
 /// table stay serial under Lance's manifest OCC, so cross-table staging has
 /// no shared state to race.
 ///
-/// 8 is a conservative default — enough to overlap S3 round-trip latency
-/// across the typical 10-30 table schemas without flooding the runtime.
-/// Override via `OMNIGRAPH_LOAD_CONCURRENCY`.
+/// The default of 8 preserves the loader's existing bound. Override it via
+/// `OMNIGRAPH_LOAD_CONCURRENCY`.
 pub(crate) const DEFAULT_STAGE_WRITE_CONCURRENCY: usize = 8;
 
 /// Resolution order: the scoped test override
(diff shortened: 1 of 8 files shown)


memmmmike:
Thanks both bots — all three points addressed in 16a1e87.

**1. `#[serial]` doesn't isolate the env mutation** (greptile + codex, agreed)

Right, and the reasoning goes further than flakiness: `#[serial]` only excludes other `#[serial]` tests, so `EnvGuard` raced every unannotated test in this binary, and `set_var` under a live multi-thread runtime violates `setenv`'s thread-safety precondition regardless.

Rather than annotate the rest of the file, the width override now rides the seam this repo already established for exactly this problem — `instrumentation::with_traversal_mode`, whose doc comment spells out the rationale ("scope-bound … process-safe … removing the need for `#[serial]`"). Added `with_stage_write_concurrency` alongside it; `stage_write_concurrency()` consults the scoped override, then the env var, then the default. `EnvGuard` and `#[serial]` are both deleted, and no `unsafe` remains in the test.

This works because staging is explicitly off-queue (it runs on the caller's task; only the commit is queued), so the task-local reaches `stage_all`.

One thing I'd have missed without writing it down: a scoped override that silently fails to apply would make the equivalence test run at the default width on *both* sides and assert nothing. There's now a unit test pinning that the override reaches the resolver, that `0` falls back rather than pinning the width to zero, and that it's gone once the future resolves.

**2. Document that the load knob now controls mutations** (greptile + codex, agreed)

`docs/dev/writes.md` described it only inside the `LoadMode::Overwrite` paragraph. Updated in both relevant places: the `stage_all` bullet now states the width is shared with the loader and why cross-table staging has no shared state to race, and the Overwrite paragraph notes the same knob tunes ordinary multi-table mutations.

**3. History-depth cost gate** (codex P1, agreed)

Fair — the equivalence test proves sameness on two shallow stores and nothing about cost. Added `multi_table_staging_is_flat_in_history` to `write_cost.rs`, in that file's existing vocabulary (`cost_harness` / `commit_many` / `assert_flat`), asserting a two-table mutation's `__manifest` and data reads stay flat from depth 10 to depth 100. Measured on local FS: depth~10 `__manifest`=11 / data=21, depth~100 `__manifest`=10 / data=19 — flat within fixture noise, and a history-proportional regression is ~10x at depth 100, so it trips immediately. That's the guard against a future change making each stage re-resolve per-table state at depth: results would stay correct and the equivalence test would stay green while the write got quadratically slower.

On the other half of that P1 — asserting the stages *overlap* — I did not add a timing or peak-inflight assertion, deliberately. Every version I could write was either timing-based (flaky in CI) or required instrumenting the production staging loop to prove a scheduling property that `.buffered(concurrency)` already gives structurally. The cost gate plus the width-seam test cover the regression I actually care about; happy to add an overlap probe if you'd rather have it explicit.

**Also:** the original commit was never `cargo fmt`'d, and CI on this branch has never run (`action_required`), so the fmt job would have failed the moment you approved it. Fixed in the same commit — `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and the `writes` / `write_cost` / lib suites (43 / 13 / 297) are all green locally on 1.97.1.

aaltshuler:
Maintainer correction pushed in 02b50092 after rebasing onto current main.

- Preserved main's deterministic pending/delete ordering and the load.between_table_stages crash-window hook.
- Added a post-rendezvous probe that proves real table-storage futures overlap: width 1 peaks at 1; width 4 reaches 2 for the two-table mutation. The semantic-equivalence assertions remain.
- Scoped the contract accurately: existing constructive datasets stage concurrently; deferred first-touch branch effects and deletes remain serial; publication is still one graph commit.
- Added current developer/user documentation and the v0.10 release note for OMNIGRAPH_LOAD_CONCURRENCY.

Local verification:
- writes: 43/43
- write_cost: 13/13
- Lance surface guards: 32/32
- staging resolver unit tests and failpoint catalog guard
- omnigraph-engine all-target Clippy with warnings denied
- formatting, documentation, AGENTS link, and workflow-pin checks

An independent final review found no remaining actionable issues.

Issue #485, opened 2026-08-11
Vector ANN is effectively brute-force: indexes are built with 1 IVF partition, and nearest() sets no probe budget


aaltshuler:
Two defects that only make sense fixed together. Reported by @FlorianSeemann in #432's comments from a production deployment (852k × 3072-dim vectors on object storage); splitting them out so they're tracked as what they are.

**1. Every vector index we build is mono-partition.** `create_vector_index` hardcodes `VectorIndexParams::ivf_flat(1, MetricType::L2)` (`table_store.rs`, the `IndexBuildSpec::Vector` staging arm). One IVF partition means the index can never prune: every ANN query reads the entire raw-vector payload. And `optimize_indices` delta-folds never re-partition, so an index built this way stays degenerate forever as the table grows. Measured impact at the reporting deployment: ~10.5 GB read per query, ~60 s on a LAN, two concurrent queries OOM'd a 24 GB host. It reproduces through both paths — fresh index creation and retrain.

**2. `nearest()` scans run with no probe budget.** The engine never calls `nprobes` (no occurrence in engine or CLI); Lance's default then probes every partition. This is why fixing the partition count alone changes nothing — measured on the same table: engine defaults **289 s**, `nprobes=20` **7.5 s**.

**Fix set** (field-tested on the reporter's patch branch, running in their production):

- Size partitions at index creation: `clamp(sqrt(non-null vectors), 1, 4096)` — never above the vector count, since k-means can't train more centroids than vectors. (Size-aware params were already flagged as a follow-up in our own code comments — `table_ops.rs`, the `vector_column_trainable` doc.)
- Pin a probe budget on `nearest()` and the rrf vector leg: default 20 (pylance's long-standing default), overridable via env.
- An `optimize --retrain-vector-index` repair verb (composes with #432's `--table` scoping): rebuilds existing trainable vector indexes through the current sizing with `replace=true` — required, because delta-folds can never fix a stale partition count.

Re-audit Lance's current `minimum_nprobes`/`maximum_nprobes` defaults at the pinned version when implementing — the measured behavior above is from the 0.8.1-era pin.

Companion visibility issue: #486.

FlorianSeemann:
We've been running a fix for this in production and can share what worked.

[...]

Issue #432, opened 2026-08-03
Feature request: table scoping for `optimize` (--table / --exclude-table)


FlorianSeemann:
### Problem

`omnigraph optimize` visits every node + edge table. When one table is operationally
un-compactable — in our case a blob-bearing `File` table (~151 GB of document blobs; since the
Lance 9.0.0-beta.15 bump removed the blob-compaction skip, optimize rewrites it) — the whole
maintenance sweep becomes unrunnable inside any reasonable operator window:

- every scheduled run dies at its timeout mid-File-rewrite, restarting that table's compaction
  from scratch on the next tick, and
- crucially, **index-coverage maintenance for every other table dies with it**. In production
  that meant fragments appended to a 751k-row vector table (`Chunk`, 3072-dim `@embed` +
  `@index` ANN) were never folded into the index. Reads stayed correct via the planner's
  brute-force flat-scan union branch, but each semantic query then pulled ~10.5 GB of raw
  vectors from object storage (~60 s per query; two concurrent queries OOM'd a 24 GB host).

The blob rewrite is also a storage-headroom hazard: compaction keeps old versions reachable
until `cleanup`, so an unscoped run transiently near-doubles the largest table in the bucket.

### Proposal

Let the operator scope the sweep:

```
omnigraph optimize --exclude-table File s3://…            # everything except the blob table
omnigraph optimize --table Chunk --table ChunkOfArtifact s3://…   # just these two
```

- Names resolve against the catalog as `node:<Type>` / `edge:<Type>` or a bare type name
  matching either; an unknown name errors (fail closed — a typo must not silently change the
  sweep), as does a scope that selects nothing.
- Scoped runs keep the exact same per-table machinery (write gates, recovery sidecar, manifest
  publish); they just visit fewer tables.
- The internal `__manifest` compaction still runs regardless of scope (cheap, in-place,
  content-preserving — and skipping it would let per-write metadata scans grow whenever the
  operator is forced to run scoped).

### Patch

We are running this in production as a minimal patch on top of `v0.8.1`:
https://github.com/FlorianSeemann/omnigraph/tree/p9/v0.8.1-optimize-table-scope
(`OptimizeScope { include, exclude }` + `optimize_tables_scoped` in
`db/omnigraph/optimize.rs`, CLI flags, and four tests in `tests/maintenance.rs` including a
scoped-reindex coverage test). Happy to port it onto `main`'s reworked optimize (the graph-wide
[...]

PR #532, opened 2026-08-18
feat(optimize): surface vector index layout so coverage answers can-it-prune

### LINKED ISSUE #486: Index coverage reports a degenerate mono-partition vector index as fully covered — the failure is invisible to monitoring
From @FlorianSeemann's 2026-08-09 report in #432, hit during a full graph rebuild: the ANN index on a fresh 853k-row × 3072-dim table came out mono-partition (see #485), so every query flat-scanned ~10 GB — while:

- `optimize --coverage-only` reported `pending_indexes: []` — complete coverage; and
- a stock `optimize` pass reported no-op — nothing detects or repairs the layout.

Coverage currently answers "does an index exist over these rows" but not "can it prune." Anyone using coverage as their health signal — which is what it's for — believes the index is fine while queries do full-payload reads.

Ask: surface partition count (and rows-per-partition, or an explicit degenerate-layout flag) in coverage/stats output, so this state is visible; the repair itself is #485's `--retrain-vector-index`.

memmmmike:
Resolves #486.

Coverage currently answers "does an index exist over these rows" but not "can it prune" — the exact blindspot from the #432 incident, where `pending_indexes: []` reported a healthy index while every `nearest()` flat-read ~10 GB.

**What this adds (reporting only; the repair remains #485):**
- `TableOptimizeStats.vector_index_layouts`: per built vector index — name, partition summary, indexed rows, worst rows-per-partition across delta segments, and a `degenerate` flag.
- The flag is computed **per segment**: Lance searches delta segments independently, so a well-partitioned later segment cannot repair a mono-partition earlier one. It fires on rows-per-partition (> 4096, generous slack over ~sqrt(N) healthy sizing), which catches both the mono-partition case and the incident's own 8-partitions-over-852k shape — partition count alone would miss the latter.
- CLI: layouts print under each table; degenerate ones loudly, with the worst segment's rows/partition. `--json` carries the full struct.
- Statistics missing the expected fields are a loud error, not a silent skip — a silent skip would recreate the exact failure mode this issue is about. `lance_surface_guards` now pins `num_indexed_rows` and `num_indexed_rows_per_delta` (aligned with `indices[]`) beside the existing `num_partitions` pin, so a Lance upgrade that changes the shape fails the guard instead of degrading the stats.

**Scope checked:** `skipped_for_drift` intentionally reports no layout (that HEAD is not manifest-accepted); `TableOptimizeStats` is not serialized by the server, so no OpenAPI drift.

**Tests:** classifier unit tests including the incident shape and the bad-segment-behind-good-segment case; an end-to-end maintenance test asserting layouts surface through `optimize()` on a real vector index; the guard pins. Maintenance suite 34/34. CLI smoke on a local graph verified both text and `--json` output, and a no-vector-column graph stays clean.

https://claude.ai/code/session_01E4S9Xpe9G7pLEv9JaoXUAb

<h3>Greptile Summary</h3>

The PR extends optimize statistics with physical vector-index layouts so callers can distinguish index coverage from actual pruning capability.
- Collects per-segment Lance partition and indexed-row statistics.
- Classifies layouts using the worst rows-per-partition segment.
- Exposes layout details through library statistics and CLI JSON/human output.
- Adds classifier, Lance-surface, and end-to-end maintenance coverage.

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph/src/table_store.rs | Parses Lance vector-index statistics into aligned per-segment partition and row counts with loud shape validation. |
| crates/omnigraph/src/db/omnigraph/table_ops.rs | Defines the public layout report and classifies degeneracy from the worst independently searched delta segment. |
| crates/omnigraph/src/db/omnigraph/optimize.rs | Threads vector-index layouts through the optimize result paths. |
| crates/omnigraph-cli/src/main.rs | Adds vector-index layout details to optimize JSON and human-readable output. |
| crates/omnigraph/tests/maintenance.rs | Verifies that optimize surfaces a real vector index's physical layout. |

</details>

```mermaid
flowchart LR
  Lance[Lance index statistics] --> Store[TableStore layout parser]
  Store --> Classifier[Per-segment degeneracy classifier]
  Classifier --> Stats[TableOptimizeStats]
  Stats --> JSON[CLI JSON output]
  Stats --> Human[CLI human output]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["chore: satisfy the fmt and clippy gates"](https://github.com/modernrelay/omnigraph/commit/9a44f04b34a762e3dbc8030357738b09cfe37eed) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54209504)</sub>

**Context used:**

- Knowledge Base — [CLI and Config](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/cli-and-config.md)

### DIFF
--- a/crates/omnigraph-cli/src/main.rs
+++ b/crates/omnigraph-cli/src/main.rs
@@ -978,6 +978,14 @@ async fn main() -> Result<()> {
                             "column": p.column,
                             "reason": p.reason,
                         })).collect::<Vec<_>>(),
+                        "vector_index_layouts": s.vector_index_layouts.iter().map(|v| serde_json::json!({
+                            "column": v.column,
+                            "index_name": v.index_name,
+                            "partitions": v.partitions,
+                            "indexed_rows": v.indexed_rows,
+                            "max_rows_per_partition": v.max_rows_per_partition,
+                            "degenerate": v.degenerate,
+                        })).collect::<Vec<_>>(),
                     })).collect::<Vec<_>>(),
                 });
                 print_json(&value)?;
@@ -996,6 +1004,19 @@ async fn main() -> Result<()> {
                     }
                     for p in &s.pending_indexes {
                         println!("    ↳ index pending on '{}': {}", p.column, p.reason);
+                    }
+                    for v in &s.vector_index_layouts {
+                        if v.degenerate {
+                            println!(
+                                "    ↳ DEGENERATE vector index on '{}': {} partition(s) over {} rows ({} rows/partition worst segment) — cannot prune; delta optimize will not re-partition (#485)",
+                                v.column, v.partitions, v.indexed_rows, v.max_rows_per_partition
+                            );
+                        } else {
+                            println!(
+                                "    ↳ vector index on '{}': {} partition(s) / {} rows",
+                                v.column, v.partitions, v.indexed_rows
+                            );
+                        }
                     }
                 }
             }
--- a/crates/omnigraph/src/db/mod.rs
+++ b/crates/omnigraph/src/db/mod.rs
@@ -12,7 +12,7 @@ pub use manifest::{Snapshot, SnapshotScanner, SnapshotTable, SubTableEntry, SubT
 pub(crate) use omnigraph::ensure_public_branch_ref;
 pub use omnigraph::{
     CleanupPolicyOptions, EXPORT_CHUNK_MAX_BYTES, ExportCut, InitOptions, MergeOutcome, Omnigraph,
-    OpenMode, PendingIndex, RepairAction, RepairClassification, RepairOptions, RepairStats,
+    OpenMode, PendingIndex, VectorIndexLayout, RepairAction, RepairClassification, RepairOptions, RepairStats,
     SchemaApplyOptions, SchemaApplyResult, SkipReason, TableCleanupStats, TableOptimizeStats,
     TableRepairStats,
 };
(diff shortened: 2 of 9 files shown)


memmmmike:
Heads-up while this waits for a CI approval: both gates would have failed on the first run, so I've fixed them in 9a44f04 (no behavior change).

- `cargo fmt --all --check`: three files this PR touches were unformatted.
- `cargo clippy --workspace --all-targets -- -D warnings`: `manual_contains` on the degenerate-partition check — `segment_partitions.iter().any(|p| *p == 0)` is now `segment_partitions.contains(&0)`.

Since CI on this branch is `action_required` and has never executed (0s), neither gate had reported, so this was invisible from the PR page. fmt, clippy, and the lib / `maintenance` / `lance_surface_guards` suites (296 / 32 / 34) are green locally on 1.97.1.

aaltshuler:
Closing this implementation while keeping the accepted problem in #486 open.

The need to distinguish index coverage from actual pruning readiness is real, but this public optimize/CLI contract needs an RFC/design pass before we ship it:

- The current stats path calls Lance 10 index_statistics(), whose default auto-migration/recompute path can create a dataset version. That violates the read-only/no-HEAD-movement boundary before OmniGraph recovery owns an effect.
- The fixed 4,096 average-rows-per-partition heuristic is not supported by pinned Lance 10 semantics or benchmark evidence, can hide partition skew, and assumes delta optimize will not repartition even though Lance may split or join partitions.
- Missing or malformed physical statistics should produce an explicit unknown/unavailable state, not make optimize itself fail.
- Per-segment facts, readiness semantics, query evidence, and repair ownership need one coherent contract rather than an optimize-only side channel.

#486 already points to the draft search-contract work. The follow-up should define a genuinely read-only readiness surface, model unknown explicitly, re-audit Lance 10 behavior, and require checked-in benchmark/query-profile evidence before exposing a classification. Thanks for surfacing the incident gap and for the implementation/test work here.

PR #544, opened 2026-08-22
perf(traversal): adaptive mode switch, limit pushdown, persisted CSR

### LINKED ISSUE #533: performance: traversal mode selection does not adapt as a multi-hop frontier grows
This came up while benchmarking Omnigraph to understand its IO and traversal costs before deciding whether an additional cache would be useful. The result looks like a planner issue that should be separated from the cache question.

## Summary

Omnigraph chooses between the indexed and CSR traversal paths before executing a variable-length traversal. For a traversal seeded by one node, the initial frontier is one row, so a high-fanout multi-hop traversal can be classified as selective. The indexed executor then continues using that path as the frontier grows, even when the CSR path is substantially cheaper.

On the IMDb graph used for this reproduction, the same four-hop traversal took about 3.6 seconds and read 255 MB on the automatically selected indexed path. Forcing `OMNIGRAPH_TRAVERSAL_MODE=csr` returned identical results in about 0.7 seconds and read 14 MB.

## Why it matters

A traversal can be selective at its first hop and dense by its later hops. The initial frontier therefore does not always predict the work that the indexed path will perform. In the example above, the indexed path rereads the edge table's endpoint columns for successive hops and materialises more than one million intermediate rows, although the query returns only 100 rows.

This can make an ordinary bounded traversal much slower and more expensive than the CSR path. A trailing `limit` does not bound the internal work in this case: the same four-hop traversal built the same intermediate result and read the same bytes with `limit 100` and `limit 100000`.

## Reproduction details

Environment:

- `modernrelay/omnigraph-server:v0.9.0`
- Current source at `426c7b9` has the same expand cost-model logic
- MinIO-backed cluster
- IMDb graph with 387,871 nodes and 2,446,964 `CoStarredWith` edges
- Graph optimized so the endpoint BTREE indexes are present

Query shape:

```text
query hop4_capped($id: String) {
    match { $p: Person
            $p.imdb_id = $id
            $p <coStarredWith>{1,4} $r }
    return { $r.imdb_id }
    limit 100
}
```

Median of five warm runs:

| Traversal | Server time | Bytes read | Largest intermediate result |
|---|---:|---:|---:|
| Automatic mode | 3,574 ms | 255.3 MB | 1,193,018 rows |
| Forced CSR | 742 ms | 14.3 MB | 190,856 rows |

The results were compared as sorted id sets and matched. The same direction of effect was observed from four different starting people.

## Source-level detail

`execute_expand` captures `wide.num_rows()` as `frontier_rows` and calls `choose_expand_mode` before entering the traversal executor. It performs a second check after opening the edge dataset to account for index coverage, but that check still happens before traversal execution.

`execute_expand_indexed` then loops over the requested hops, updating its local frontiers and issuing one or more indexed edge scans per hop. It does not call the cost model again as those frontiers grow. The hard frontier ceiling is therefore also an initial-dispatch ceiling, not a bound that can force a mode change during execution.

The current documentation describes the ceilings as applying to initial dispatch and describes the estimate as approximately `hops × frontier × fanout`. The observed behaviour is consistent with that implementation, but the initial frontier is a poor estimate for a single-seed traversal with rapid fan-out.

## Question

Could the traversal planner account for frontier growth during execution, or otherwise avoid committing a high-fanout multi-hop traversal to the indexed path based only on its initial frontier? The appropriate change may be an adaptive switch, a more conservative estimate, a bounded indexed phase, or another approach.

For comparison, SQL Server's adaptive joins retain alternative join strategies and defer the choice until runtime rows cross a threshold. This is not intended as a proposed implementation for Omnigraph, but it is an example of a mature query engine making a related runtime decision based on observed cardinality: <https://learn.microsoft.com/en-us/sql/relational-databases/performance/joins>.

This report is intended to establish the performance issue without prescribing the implementation.

azimafroozeh:
## What & why

Closes #533. The traversal-mode chooser ran once before hop 1, so a single-seed high-fanout traversal (frontier 1 at dispatch, 92,889 by hop 4 on the reported shape) committed to the indexed path and stayed there. This PR reworks the Expand execution path around that bug and ships four improvements; every measured result is collected in the results section below.

1. Adaptive mode switch: the cost decision reruns at every indexed hop with the observed frontier (projected forward by the observed growth ratio; the frontier ceiling is now an execution bound, and undirected traversals are priced at their true two probes per hop). An outgrown traversal switches to CSR mid-flight, carrying its BFS state, no restart.
2. Limit pushdown: an unordered, aggregate-free `limit` on a final filterless Expand stops the traversal at the cap, with an uncapped rerun if hydration drops below it.
3. Unified BFS core: the duplicated indexed and CSR executors become one loop over a pluggable neighbor source; every traversal contract (emission, dedup, hop gating, self-edge, cross-type) is written once, and the mid-flight switch is just a source swap. Pure-path overhead is within noise.
4. Persisted graph index: `optimize` writes the CSR/CSC plus dictionaries to `__graph_index/csr-current.bin` (binary sections, per-edge-dataset identity stamps, payload sha256); cold traversal builds load it with one GET instead of scanning every edge table.

### Measured results, before this PR vs after

Environment: local NVMe, release build, medians of warm runs, both legs from the same instrument on the same machine; the before leg is the pristine merge base. The instrument is a drop-in example (the #533 reproduction extended with scale arguments), not part of this PR. Query legs use the #533 graph shape: 388k nodes, 2,519,918 edges, seed degree 238, per-hop frontier 1 / 238 / 5,418 / 92,889 against the 1,024 ceiling.

What the engine does on its own (auto mode):

| Scenario | Before this PR | After this PR | Speedup | What changed |
|---|---:|---:|---:|---|
| The issue's query: 4 hops, undirected, `limit 100` | 750.0 ms | 8.6 ms | **87x** | limit pushdown stops the traversal at hop 1 |
| Same query, uncapped (250,442 results) | ~750 ms (auto tracked forced-indexed, 753.8 ms) | 247.8 ms | **3.0x** | the mid-flight switch moves auto onto the CSR path |
| Cold CSR acquisition, 25M-edge store | 1.95 s (edge-table scan) | 229 ms (one GET of the 240 MiB artifact) | **8.5x** | persisted graph index |

Sanity: results are identical in every mode and every scenario (100 and 250,442 rows respectively), the after-auto uncapped time matches forced CSR (247.8 vs 247.9 ms), and the forced modes themselves are unchanged within noise, so the wins come from better decisions, not a changed engine core.

Format decision behind the artifact (not a before/after: upstream never had an artifact): both encodings measured over identical in-memory data at 100M edges (1,060 MiB raw). Binary ships because it decodes 1.9x faster than JSON+base64 (0.61 s vs 1.17 s, paid on every cold load) and is 21% smaller (1,061 vs 1,339 MiB, paid on every cold GET). A digest-free floor variant (0.30 s) was measured and rejected: about half of binary decode is the sha256, kept unconditionally because raw arrays have no syntax to fail on and a flipped byte would otherwise become silently wrong topology.

## Backing issue / RFC

- [x] Fixes an **accepted** issue: Closes #533

## Checklist

- [x] Change is focused (one rework of the Expand execution path; the pieces share the unified core)
- [x] Tests added/updated (`traversal_adaptive` suite: 8, incl. switch equality directed/undirected, min-hops across the switch, capped-limit subset validity, artifact write/corruption/staleness; indexed/CSR equivalence battery 13 now exercises the unified core; chooser units 15; persist units 22 incl. the crafted-artifact class)
- [x] Public docs updated (`docs/user/queries/index.md`, `docs/user/reference/constants.md`, `docs/user/operations/maintenance.md`, `docs/user/concepts/storage.md`, `docs/dev/invariants.md`)
- [x] Reviewed against [docs/dev/invariants.md](../blob/main/docs/dev/invariants.md): no Hard Invariant weakened; `__graph_index/` is derived, regenerable, optimize-only-written, never authoritative

## Local verification

- `cargo test --workspace --no-fail-fast`: 80 suites ok; the one failure is the pre-existing local-environment `external_blob_file_policy_rejects_special_files`
- `cargo clippy --workspace --all-targets` clean; `cargo fmt --all` applied
- vocabulary guard: `user-docs` and `openapi` clean; `rust-string` inventory updated for the executor rename and persist.rs (rows reviewed)

## Notes for reviewers

- Semantics change: unordered, aggregate-free `limit n` now returns an arbitrary valid n-subset instead of the previous deterministic slice (count exact; ordered/aggregated queries untouched). Multi-seed CSR row order also changed (hop-major); unordered order is now explicitly uncontracted. Both documented.
- `__graph_index/` is a new top-level store prefix: one derived object, `optimize` is the sole writer, queries only read, every failure or staleness falls open to the in-memory scan build; no GC yet.
- The loader validates beyond the digest (offsets cover the keyed dictionary, targets bounded by the opposite one, allocation clamped, every stamped edge must carry its adjacency), so an internally inconsistent artifact is rejected fail-open, never a panic or dispatch error.
- New byte primitives `read_bytes_if_exists_bounded` / `write_bytes` on both StorageAdapter traits; forbidden_apis registers the persist write as PhysicalOnly (derived, regenerable, never graph-visible).
- The artifact's stamps must stay in lockstep with the in-memory graph-index cache key (cross-referenced in both files); on e_tag-less local FS it inherits that key's documented branch-ref ABA residual.
- Known deferred costs, follow-up with a shared fix (header-first ranged reads): a stale artifact is rejected only after a full-object GET, and a scoped load caches the full-catalog index under its scoped key.

<h3>Greptile Summary</h3>

The PR unifies traversal execution, adds adaptive indexed-to-CSR switching and safe limit pushdown, and persists a validated topology artifact during optimization.
- Re-evaluates traversal cost at each indexed hop and carries BFS state into CSR mode.
- Pushes eligible unordered limits into final expansions, rerunning uncapped when hydration underfills.
- Adds bounded binary storage operations and a digest- and identity-validated persisted graph index.
- Reserves Lance search-output property names for newly parsed schemas while retaining compatibility with stored catalogs.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph/src/exec/query.rs | Unifies indexed and CSR traversal, preserves BFS state during adaptive switching, and safely pushes eligible final limits with an uncapped hydration fallback. |
| crates/omnigraph/src/graph_index/mod.rs | Extends graph-index acquisition and identity handling to support persisted topology without changing snapshot semantics. |
| crates/omnigraph/src/graph_index/persist.rs | Implements bounded, checksummed artifact serialization and fail-open loading with structural and per-edge identity validation. |
| crates/omnigraph/src/db/omnigraph/optimize.rs | Persists the derived graph index after optimization while retaining the main-branch maintenance gate. |
| crates/omnigraph-storage/src/lib.rs | Adds bounded binary reads and atomic-visibility binary writes across supported object-storage backends. |
| crates/omnigraph/src/runtime_cache.rs | Adds persisted-artifact shelving and negative caching while revalidating entries against each requested snapshot. |
| crates/omnigraph-compiler/src/schema/parser.rs | Rejects new declarations that collide with Lance search ranking output columns without invalidating existing stored catalogs. |
| crates/omnigraph-dst/src/harness.rs | Integrates binary storage calls into fault, completion, and durable-write accounting, with explicit exclusions for text-only fault modes. |

</details>

```mermaid
flowchart LR
  Q[Traversal query] --> I[Indexed BFS]
  I --> C{Re-evaluate cost per hop}
  C -->|Indexed remains cheaper| I
  C -->|Frontier outgrows indexed path| A[Acquire CSR]
  A --> P{Fresh persisted artifact?}
  P -->|Yes| L[Load validated CSR/CSC]
  P -->|No or invalid| B[Build CSR/CSC from snapshot]
  L --> S[Translate BFS state]
  B --> S
  S --> R[Continue unified BFS]
  R --> H[Hydrate destinations]
  H --> U{Pushed limit underfilled?}
  U -->|Yes| X[Rerun uncapped]
  U -->|No| O[Project and apply final limit]
  X --> O
```
</details>

<sub>Reviews (7): Last reviewed commit: ["review"](https://github.com/modernrelay/omnigraph/commit/bb2119bc6eada7bf6527b53950b3fcac761ee82c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55884791)</sub>

- Knowledge Base — [Query execution and projection](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/query-execution.md)
- Knowledge Base — [Graph indexing and search](https://app.greptile.com/modernrelay/-/custom-context/knowledge-base/modernrelay/omnigraph/-/docs/graph-indexing-and-search.md)
</details>


gordonmurray:
Thanks, Azim. As requested, I repeated the #533 workload against the current PR head, d20c23a, using the preserved IMDb and MinIO graph.

Nice work. The reported workload now behaves much better:

- The capped four-hop query improved from 2.32 seconds to 98 ms. It returned the same 100 rows and checksum.
- The 100,000-row traversal improved from 2.47 seconds to 408 ms. It returned the same checksum.
- Forced indexed mode took 1.20 seconds. Forced CSR took 408 ms, which closely matched automatic mode.
- The logs confirmed an indexed start and a switch to CSR at hop 2, with a frontier of 499.
- The focused traversal_adaptive suite passed all eight tests.

I found no correctness mismatch in these runs.

The original #533 results used an older Omnigraph build. Therefore, I cannot attribute the full end-to-end improvement to this PR alone.

I also compared traversal modes on the same PR build. Automatic mode took 408 ms, while forced indexed mode took 1.20 seconds. This comparison shows that the adaptive switch provides a substantial improvement.

I did not run optimize on the preserved graph. Therefore, this run did not exercise the persisted CSR artifact against MinIO.

While reviewing that new path, I noticed one possible concern. The artifact contains the full catalog, but the runtime cache keys entries by query edge scope.

Different query scopes can therefore retain several decoded copies of the full artifact. I did not observe increased memory use during this run.

Could you confirm whether this cache behavior is intentional? An RSS measurement across several edge scopes would help assess it.

Based on the measured behavior, I am comfortable with the adaptive switch and limit pushdown. They solve #533 for this workload.

My remaining merge question applies only to persisted CSR, which I did not exercise against MinIO.

I would also like to see the full workspace and RustFS suites pass after the rebase.

crates/omnigraph/src/exec/query.rs
```diff
@@ -1808,85 +2183,176 @@ async fn execute_expand_indexed(
...
 
-        // Advance each source row's frontier independently (dense ids).
+        // Advance each source row's frontier independently (dense ids). The
+        // emission contract is identical across sources: a self-edge is a
+        // valid destination that reaches nothing new — emit it without
+        // entering the frontier, so the seeded-source `visited` pre-mark
+        // prunes only multi-hop cycle returns (same-type only: cross-type
+        // dense ids live in different namespaces, where node == neighbor is
+        // meaningless for Csr and collision-prone for Indexed).
         for i in 0..n {
```
aaltshuler, line 2708: ### [P1] Preserve incoming search rank through multi-hop expansion

The shared BFS processes every source at hop 1 before processing hop 2, whereas the old CSR path completed each source in input order. This interleaves a lower-ranked ANN/BM25 source ahead of later-hop results from a higher-ranked source. `execute_query` skips final sorting for search-ordered queries, so the final `LIMIT` returns the wrong top results even though search queries correctly disable traversal limit pushdown.

Reproduced by extending `search.rs::nearest_rank_survives_bound_edge_fanout` with a `sink -> rank-3` edge, then querying `$d rankedLink{1,2} $target`, ordered by `nearest($d.embedding, $q)`, with `limit 2`. Both automatic and forced-CSR execution return `(rank-1, sink), (rank-2, sink)` on this head; the identical diagnostic passes on the merge base with `(rank-1, sink), (rank-1, rank-3)`.

Please preserve the incoming source rank when materializing traversal results and restore that ordering before the final limit. Add multi-hop ranked-search coverage; the existing one-hop fanout assertion does not catch this regression.

crates/omnigraph/src/graph_index/persist.rs
```diff
@@ -0,0 +1,770 @@
...
+                    "graph index artifact does not cover edge '{edge_name}'"
+                ))
+            })?;
+        if *stamped != expected {
+            return Err(OmniError::manifest(format!(
+                "graph index artifact is stale for edge '{edge_name}'"
+            )));
+        }
+    }
+    let digest = base64::engine::general_purpose::STANDARD.encode(sha2::Sha256::digest(payload));
```
aaltshuler, line 402: ### [P1] Include interpretation metadata in artifact integrity checking

The digest covers only `payload`, while the unchecked header determines which edge type each adjacency section belongs to. Structural checks cannot detect all incorrect interpretations of otherwise valid arrays.

Reproduced by extending `traversal_adaptive.rs::corrupt_artifact_falls_back_to_the_scan_build` with two same-endpoint edge types, `Knows` and `Likes`. Exchanging only their adjacency section labels leaves the payload, its checksum, header length, and freshness stamps unchanged. The loader nevertheless accepts the artifact: a `Knows` traversal returns the lone `Likes` destination (`q1`) instead of its 49 actual neighbors. The underlying graph data is intact; the derived artifact changes the logical answer instead of triggering fallback.

Please checksum the complete versioned envelope, including metadata as well as payload, and validate the header's structural relationships before accepting it. Retain a metadata-corruption regression in addition to the existing truncated-payload coverage.

crates/omnigraph/src/runtime_cache.rs
```diff
@@ -81,8 +82,11 @@ impl RuntimeCache {
             }
         }
 
-        crate::instrumentation::record_graph_build(edge_types.len());
-        let index = Arc::new(GraphIndex::build(&resolved.snapshot, edge_types).await?);
+        // The graph-build probe fires inside `GraphIndex::build` itself, so a
+        // persisted-artifact load is never counted as a build.
+        let index = Arc::new(
```
aaltshuler: ### [P1] Share the full persisted artifact across scoped cache entries

The key remains scoped to the requested edge set, but `GraphIndex::load_or_build` now loads and decodes the entire persisted catalog on every miss. Sequential `{Knows}` and `{WorksAt}` requests therefore retain two independent full-graph allocations, not two small scoped indexes. The eight-entry LRU can retain eight complete copies, with additional transient allocation during loading, and each distinct scope repeats the full artifact read/decode. This is a substantial memory and I/O regression for large optimized graphs.

Confirmed with a review-only runtime-cache test using the existing standard graph fixture: both requests return an index containing both edge types, but their `Arc`s are distinct. The test proves duplication; large-scale OOM was not exercised.

Please admit/decode the persisted artifact once per graph handle and share its allocation across compatible scopes, retaining per-request freshness checks. Keep scan-built fallbacks scoped when the artifact cannot serve a request, and bound concurrent admission so simultaneous misses do not duplicate the full decode.

aaltshuler (commented):
Reviewed the exact PR head, d20c23ab, against merge base 62a9c3fe. Three confirmed findings below need addressing before merge.

Verification: 77 existing tests passed across the focused traversal/aggregation/equivalence suites, persisted-index unit tests, and runtime-cache unit tests. Additional review-only diagnostic extensions reproduced the three findings; the search-ranking diagnostic passes on the merge base and fails on this head. These diagnostic extensions are not part of the submitted PR. Full-workspace and AWS qualification were not performed.

crates/omnigraph/src/exec/query.rs
azimafroozeh: Confirmed and fixed. A final sort would work but pays O(R log R), so it is a top-k: Lance already emits _distance/_score and hydration copies them onto every traversal row, so execute_query replaces the sort skip (exec/query.rs:493) with a top-k on that column, limit as fetch, O(R log n). Emission order stops being load-bearing. Search-ordered aggregates keep today's behavior; equal-score ties become id-ordered, deterministic, documented; secondary order keys after the search function are now honored too, and _distance/_score become reserved property names so user data can never shadow the ranking. Multi-hop ranked coverage added per this repro, plus a distinct-score bm25 golden and a secondary-key regression.

crates/omnigraph/src/graph_index/persist.rs
azimafroozeh: Confirmed and fixed: the meaning moved inside the digest. Each payload section opens with its own prelude (edge type, orientation, identity stamp) inside the digested bytes; the loader cross-checks the header's section map against them, so the header becomes an advisory index: a lie causes rejection, never wrong topology. Version bumped to v2; old artifacts fall open to the scan build (release-noted, one-time rebuild until the next optimize). Label-swap and stamp-swap regressions beside the truncation test, plus a whole-index round-trip oracle.

crates/omnigraph/src/runtime_cache.rs
azimafroozeh: Confirmed; deferring it was the wrong call. Implemented in your shape: the handle slot is keyed by artifact identity (stamps + payload digest), freshness stays per-request, admission is single-flight, and a negative cache remembers failed attempts per scope key so a stale artifact never queues misses behind repeated doomed GETs. A regression test asserts one shared allocation across two scopes.

Plan for the branch: rebase with conflict resolution, then full workspace and RustFS suites on the new head.

PR #555, opened 2026-08-25
feat(bench): add declarative harness foundation

aaltshuler:
## What and why

Introduces the reviewed declarative foundation for `omnigraph-bench`: strict versioned YAML case and suite definitions, semantic validation, canonical experiment identity, stable point IDs, and an execution-free planning CLI.

A benchmark case owns the complete experiment identity. A suite selects checked-in cases and sets repetition quantity. The full SHA-256 of the canonical typed identity is the comparison key; readable point names are display-only.

## Cache-condition contract

Cache state is modeled as a structured condition instead of one ambiguous cold or warm label:

- `process`: `fresh-per-repetition`
- `engine`: `preparation-only`, `warmed-by-program`, or `reopened-after-program`
- `page_cache`: `uncontrolled` or `program-conditioned`
- `program`: `none` or a named versioned read program
- `iterations`: bounded and identity-bearing

Only three complete tuples validate: process-cold, warm, and post-reopen. Process-cold means a fresh worker with no declared warm-up program; it does not claim an empty OS page cache or cold storage. Program-conditioned records the treatment that ran, not a residency proof. True page-cache-cold and storage-cold experiments are intentionally unrepresentable until a platform-specific control and witness exist.

## Scope

- Adds typed and bounded `case-v1` and `suite-v1` schemas with cross-field validation.
- Adds `case validate`, `case list`, `suite validate`, and `suite plan`, including JSON output.
- Adds a checked-in branch-merge smoke case and local suite.
- Captures exact backend identity, cache condition, reset mode, fixture builder, and structured index inventory.
- Updates RFC 0039 and benchmark documentation around repetition independence, cache claims, and logical versus physical fixture identity.

This slice deliberately does not execute workloads, persist measurements, or orchestrate AWS. Those land behind the stabilized `ResolvedRun` contract in separate PRs.

## Relationship to #537

This is the clean successor to the configuration, identity, and CLI layer of #537. The engine timing seam from that PR already landed through #551. Selected runtime mechanics are adapted in #556, while the drifting repetition loop, lossy point key, record and noise-floor model, and weak verification are not carried forward.

## Validation

- `cargo test -p omnigraph-bench --locked`
- `cargo clippy -p omnigraph-bench --all-targets --locked -- -D warnings -W clippy::dbg_macro`
- `cargo fmt --all --check`
- `bash scripts/check-agents-md.sh`
- `python3 scripts/check-docs.py`
- `python3 scripts/check-workflow-action-pins.py`

## Next slices

[...]

PR #537, opened 2026-08-19
feat(bench): RFC 0039 benchmark harness, branch merge as first workload

azimafroozeh:
## What & why

Adds `omnigraph-bench`, the harness for the end-to-end benchmark (RFC 0039). It owns the general machinery: frozen validated fixtures, append-only schema-validated run records, an A/A noise floor with rule-7 claim labeling, and terminal plus self-contained HTML reporting. Branch merge is the first workload wired through it, measuring wall-clock with 14-phase attribution and storage-call counts per run.

First result from the harness: on the large fixture (8 tables, 100k rows, delta 5000) the three-way row walk is ~90% of merge wall-clock, and merge cost tracks table size, not delta size. The walk had no phase bucket before; the new `TableWalk` phase records it directly.

## Backing issue / RFC

- [x] Is an RFC PR, or implements a **merged** RFC: docs/rfcs/0039-end-to-end-benchmark.md (RFC PR #535; this PR lands after it merges)

## Checklist

- [x] Change is focused (one logical change: the harness plus the minimal engine seam it reads)
- [x] Tests added/updated for behavior changes (48 unit tests in the crate; engine drift guards extended for the 14th phase)
- [x] Public docs updated if user-facing surface changed (crate README, AGENTS.md crate list and topic map; internal instrument, no public API surface)
- [x] Reviewed against docs/dev/invariants.md — no Hard Invariant weakened, no deny-list item hit (engine change is one observational enum variant plus two timing spans, no-op unless a probe is installed)

## Local verification

- `cargo fmt --all --check` — clean
- `cargo clippy -p omnigraph-bench --all-targets -- -D warnings` — clean
- `cargo clippy -p omnigraph-engine --lib -- -D warnings` — clean
- `cargo check -p omnigraph-engine --benches` — clean
- `cargo test -p omnigraph-bench` — 48 passed, 0 failed
- `bash scripts/check-agents-md.sh` — OK (78 links, 72 docs)
- `python3 scripts/generate_merge_bench_fixtures.py` — builds the release binary and the three v1 fixtures end to end

## Notes for reviewers

- Engine footprint: `MergeTimingPhase::TableWalk` plus two recording spans around `stage_streaming_table_merge`. Production pays one unset task-local check.
- Rule 6: nothing here gates anything, at any CI stage. Workspace member, `publish = false`.
[...]

PR #556, opened 2026-08-26
feat(bench): execute identical-state local merge runs

aaltshuler:
## What and why

Executes validated `ResolvedRun` plans from declarative YAML through a deterministic local branch-merge runner. Builder v2 creates an equal mix of node and edge tables, connects edges in a deterministic ring, and diverges edge state on two branches. The already-diverged graph fixture is built and validated in a watchdog-bounded process group, frozen once, and restored for every repetition from a never-opened APFS clonefile template at the same stable path.

Every sample runs in a fresh SHA-attested release worker through the versioned `Request -> Ready -> Begin -> Settled -> Complete` protocol. The parent owns process containment, independent watchdogs, bounded output capture, peak-RSS observation, sample admission, and cleanup.

## Stack

This PR is stacked on #555, which owns the case and suite schema, cache-condition grammar, planning, and point identity. Merge #555 first; GitHub will then retarget this PR to `main`.

This supersedes the valid local-execution portion of #537 while preserving the reviewed declarative interface.

## Cache-condition execution

The runner dispatches from the complete validated cache tuple:

- Process-cold: a fresh worker, no declared warm-up program, and measurement on the prepared handle.
- Warm: run the named versioned read program for the declared iterations, then measure on the same handle.
- Post-reopen: run the named program, drop and reopen the engine handle, then measure.

Process-cold is deliberately narrow. Engine open and protected-head capture happen before `Begin`, and the OS page cache remains uncontrolled. The runner makes no page-cache-cold, device-cache-cold, or object-store-cache-cold claim. Unsupported tuples fail closed.

## Benchmark-validity guarantees

- The fixture owns all starting state; repetitions own no persistent history.
- Fixture verification checks node content, edge IDs and endpoints, branch effects, and untouched graph state across source, target, and main.
- Fixture construction itself is a bounded child process. Failed or partial builds quarantine their workspace.
- Every public repetition is a distinct fresh worker process.
- Reset and pre-timer witnesses traverse metadata only after the initial physical digest.
- A preparation firewall permits only the balanced local capability probe before `Begin`.
- After timing, the worker verifies exact content across every table on target, source, and main. The parent independently derives and admits identity, elapsed time, digest, merge outcome, TableWalk route, and declared table/total-row counts; it does not claim to reread the content bytes.
- Runner-v1 deliberately performs O(store) verification outside the measured window. Receipt-based O(delta) certification starts with versioned-S3 reset; a future DST oracle is additive and never replaces per-repetition probes.
- The counting wrapper overrides vectored reads, preserving one logical `get_ranges` operation rather than reshaping it into scalar reads. These remain logical store calls, not physical requests or cost evidence.
- Failed or timed-out workers must prove direct-child reap, an empty process group, and clean bounded stdio closure before cleanup.
- Worker evidence includes exact executable SHA-256, build/source facts, effective Lance memory-pool configuration, and parent-observed peak RSS.
- A null operation deadline means no declared measurement deadline; the supervisor still retains an independent safety watchdog.
- Diagnostic JSON and plain output include the exact cache condition used. Durable telemetry remains a later slice.

## Validation

- `cargo test -p omnigraph-bench --locked` — 109 library tests and 9 CLI tests
- `cargo test -p omnigraph-engine --test benchmark_scenario_contract --locked` — 7 source-guard tests
- `cargo clippy -p omnigraph-bench --all-targets --locked -- -D warnings -W clippy::dbg_macro`
- `cargo fmt --all --check`
- `python3 scripts/check-docs.py`
- `bash scripts/check-agents-md.sh`
- `python3 scripts/check-workflow-action-pins.py`

Coverage includes complete-tuple dispatch, process isolation with distinct worker PIDs, bounded fixture construction, warm and post-reopen execution, optional-deadline supervision, protected-head preparation failures, transparent vectored-read counting, parent/child clock agreement, peak RSS, diagnostic cache evidence, protocol corruption, timeout containment, portable reset proofs, fixture restoration, and exact graph verification.

## Not in this PR

Durable JSON record schemas, database projection, A/A floors and diff semantics, S3 versioned reset, server-mode execution, platform-controlled page-cache eviction, storage-cold claims, and AWS orchestration remain separate reviewable slices.

<h3>Greptile Summary</h3>

The PR extends the declarative benchmark harness with deterministic local branch-merge execution.
- Builds and validates balanced node/edge fixtures with deterministic branch divergence.
- Restores identical APFS clonefile state for each fresh worker repetition.
- Adds versioned worker supervision, bounded watchdogs, process containment, logical I/O accounting, and exact post-merge verification.
- Exposes local suite execution and updates benchmark configuration and documentation.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph-bench/src/runner.rs | Coordinates fixture ownership, per-repetition restoration, worker execution, sample admission, and cleanup. |
| crates/omnigraph-bench/src/supervisor.rs | Implements bounded worker protocol supervision, process-group containment, output capture, and peak-RSS observation. |
| crates/omnigraph-bench/src/worker.rs | Implements the private repetition-worker endpoint and its protocol lifecycle. |
| crates/omnigraph-bench/src/reset.rs | Adds physical and metadata witnesses plus forced APFS clonefile freezing and restoration. |
| crates/omnigraph-bench/src/fixture_worker.rs | Moves fixture construction and freezing into a watchdog-bounded child process. |
| crates/omnigraph-bench/src/branch_merge.rs | Builds deterministic balanced graph fixtures, applies branch divergence, and verifies exact merge results. |
| crates/omnigraph-bench/src/worker_protocol.rs | Defines the versioned parent-worker handshake and executable attestation. |
| crates/omnigraph-bench/src/preparation.rs | Restricts pre-measurement writes to the expected balanced capability probe. |
| crates/omnigraph-bench/src/counting.rs | Adds transparent logical object-store operation accounting. |
| crates/omnigraph-bench/src/environment.rs | Validates the actual macOS scratch volume against declared filesystem and storage properties. |
| crates/omnigraph-bench/src/bin/omnigraph-bench.rs | Adds suite execution and private worker entry points to the CLI. |
| crates/omnigraph-bench/src/case.rs | Updates validation and identity semantics for balanced builder-v2 fixtures and cache execution modes. |

</details>

```mermaid
sequenceDiagram
    participant CLI as Suite runner
    participant F as Fixture worker
    participant T as Frozen template
    participant W as Repetition worker
    participant E as OmniGraph engine
    CLI->>F: Build validated diverged fixture
    F->>E: Initialize and verify branches
    F->>T: Freeze APFS clonefile template
    F-->>CLI: Attested fixture handoff
    loop Each repetition
        CLI->>T: Restore stable active path
        CLI->>W: Request
        W->>E: Open and prepare cache state
        W-->>CLI: Ready
        CLI->>W: Begin
        W->>E: Merge source into target
        W-->>CLI: Settled
        W->>E: Verify exact graph content
        W-->>CLI: Complete
        CLI->>CLI: Admit sample and clean active tree
    end
```
</details>

<sub>Reviews (8): Last reviewed commit: ["merge: advance benchmark runner to main"](https://github.com/modernrelay/omnigraph/commit/5546ced3d9f400fc83c86869a982cf0dd7223262) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56909775)</sub>


aaltshuler (commented):
Validated four runner-level findings against this PR head. The unbounded host probes are blocking under the bounded-failure invariant. The worker stage mapping, df parser, and profile-specific CLI test are confirmed fail-closed or test-robustness defects.

crates/omnigraph-bench/src/environment.rs
```diff
@@ -0,0 +1,240 @@
...
+            "runner-v1 cannot map diskutil protocol `{protocol}` to a declared storage class"
+        ))
+    }
+}
+
+#[cfg(target_os = "macos")]
+fn command_output(program: &str, arguments: &[&str]) -> Result<String, String> {
+    let output = Command::new(program)
+        .args(arguments)
+        .output()
```
aaltshuler: **[P1] Bound every environment probe.** `Command::output()` has no timeout, kill, or reap path and buffers both pipes before the 64 KiB check. This runs before repetition supervision, so a wedged `df` or `diskutil` can hang the whole benchmark indefinitely. Please use a deadline-driven spawn/poll/kill/reap helper with bounded concurrent capture.

crates/omnigraph-bench/src/environment.rs
```diff
@@ -0,0 +1,240 @@
...
+    available_bytes: u64,
+}
+
+#[cfg(target_os = "macos")]
+fn parse_df_kilobytes(output: &str) -> Result<MountEvidence, String> {
+    let line = output
+        .lines()
+        .rfind(|line| !line.trim().is_empty())
+        .ok_or_else(|| "df output did not contain a mounted-volume row".to_string())?;
+    let fields = line.split_whitespace().collect::<Vec<_>>();
```
aaltshuler: **[P2] Do not parse this row by whitespace position.** A filesystem identifier containing whitespace shifts the numeric columns, and joining `fields[5..]` irreversibly collapses repeated spaces in a valid APFS mount path. On the supported macOS envelope, use structured `df --libxo=json -Pk` output or a native filesystem API and fail closed on malformed data.

crates/omnigraph-bench/src/worker.rs
```diff
@@ -0,0 +1,373 @@
...
+        | "worker_attestation_failed"
+        | "worker_case_invalid"
+        | "worker_identity_mismatch" => WorkerStageV1::Bootstrap,
+        "pre_measurement_write_detected"
+        | "pre_measurement_state_mismatch"
+        | "warmth_failed"
+        | "engine_open_failed"
+        | "storage_open_failed" => WorkerStageV1::Prepare,
+        "merge_failed" | "merge_deadline_exceeded" | "duration_overflow" => WorkerStageV1::Measure,
+        "verification_failed" | "vacuous_merge" | "missing_table_walk_phase" => {
```
aaltshuler: **[P2] `verification_failed` has two lifecycle meanings here.** `capture_protected_branch_heads` emits it before `Ready`, but this mapping labels every occurrence `Verify`; the supervisor correctly rejects a pre-Ready `Verify` frame and replaces the useful cause with `worker_protocol_error`. Give pre-measurement head capture a distinct error code mapped to `Prepare`, while retaining this mapping for post-measurement verification.

crates/omnigraph-bench/tests/cli.rs
```diff
@@ -140,6 +140,54 @@ fn missing_case_selector_fails_instead_of_running_the_full_suite() {
         .stderr(predicate::str::contains("does-not-exist"));
 }
 
+#[test]
```
aaltshuler, line 168: **[P2] Gate this debug-only expectation.** Under `cargo test --release`, the release guard succeeds and this test proceeds into real fixture setup instead of returning `release_build_required`; on a supported Mac that can launch the 800k-row, five-repetition benchmark before failing the assertion. Add `#[cfg(debug_assertions)]` and keep release attestation coverage in a cheap unit seam.

azimafroozeh:
**[P0]** CountingStore doesn't override get_ranges, so the trait default splits Lance's vectored reads into per-range gets inside the measured window. The instrument reshapes the I/O it exists to observe and inflates the counts, so every number before the fix is wrong. Same trap the file already fixes for rename_opts, inherited from #537. Fix: override it, count one call, forward the batch; add `#[warn(clippy::missing_trait_methods)]` on the impl and one test that the wrapper reaches the target's get_ranges.

**[P0]** Cold is broken in the spec and missing in the code, and cold is the number we sell. The RFC says cold means a fresh process; that's wrong, a fresh process still reads the OS page cache, so the runner rightly refuses it. But cold is the first query after a deploy, the worst-latency moment of an object-storage engine, and the cell where the cache work must prove its win; until it exists, every number here shows the engine at its warmest. Two steps: redefine cold in the RFC by proof instead of method (proven page-cache eviction per rep, or refuse), riding the deadline none edit; then a small slice implements it, per-platform eviction behind a canary-read witness.

**[P1]** The unbounded-wait finding generalizes: fixture build has no watchdog. A wedged engine in initialize_local_fixture hangs the sequential suite and pins the workspace, and the probe helper won't cover it since these are in-process awaits, not a child to kill. Either a timeout that quarantines, or an explicit scope-out in the module doc.

**[P1]** Full three-branch verification replaced the letter's receipt certification: stronger, but O(store) per rep instead of O(delta). Right call for runner-v1, but it needs one recorded line: runner-v1 ships full verification; receipt certification activates with the S3 slice, where versioning provides the receipt; the DST oracle (#527, now merged) can wire as its own follow-up.

**[P2]** containment_proven (runner.rs:878) treats missing evidence as proof: an error with no child-process evidence passes the gate, and the gate decides delete vs quarantine of the active tree. The dangerous branch is the default when nothing is known, and worker_pid_overflow already attaches nothing. Invert it: no evidence, no deletion, quarantine.

**[P2]** SUT identity, before the record format freezes: no engine git sha or dirty flag in BuildEvidence; LANCE_MEM_POOL_SIZE is 1GB under cargo run and 100MB for a pinned binary with identical evidence; per-rep peak RSS was lost with the try_wait reap.

**[P2]** split_delta and the case.rs capacity check compute the same split twice with no shared function or equality test; changing the ratio in one silently breaks the other's proof.

**[P3]** Child-reported elapsed_us is never bounded by the supervisor's own clock; one inequality closes it.

**[P3]** Three sentences overstate the code: the crate README says finalize failures quarantine (with proven containment they delete; the benchmarks README is correct), the PR body says the parent independently checks exact content (it admits the worker's attestation booleans), and "must prove containment before cleanup" is the P2 above.

**[P3]** Two more config-sensitive tests: symlinks_and_special_files binds a Unix socket, which fails deterministically in sandboxes, and the clonefile tests assume an APFS tempdir.

**[P3]** After rebasing over #527, run its new source guards against this crate; the stack predates them.

**[P4]** Diagnostics: "state" for a shape-only check (runner.rs:1096); "micro profile" in an error the README forbids encoding (runner.rs:414); Debug variant names where the YAML wants kebab-case (environment.rs, storage_name sits unused); a prescribed "history recipe" that doesn't exist (branch_merge.rs:862); admission failures print observed values but never expected.

**[P5]** Dead code: the timer check at branch_merge.rs:295 is unreachable; four error codes mislabel as Finalize via the catch-all.

aaltshuler:
Implemented the cache-condition execution follow-up on top of #555.

- Every public repetition already launches a distinct fresh release worker; a supervisor test now proves distinct PIDs.
- Process-cold skips the warm-up program and measures on the prepared handle.
- Warm executes the declared versioned program and measures on the same handle.
- Post-reopen executes that program, drops and reopens the handle, then measures.
- Dispatch matches the complete tuple and refuses anything outside the schema contract.
- Result JSON and plain diagnostics include the exact cache condition.
- Tests no longer pretend that a private in-process helper proves process-cold.
- Docs state that engine open and protected-head capture happen before Begin and that OS and storage caches remain uncontrolled.
- Null operation deadlines are supported independently from the supervisor safety watchdog.
- Pre-Ready protected-head failures now retain the Prepare stage.

This resolves the cold ambiguity without manufacturing eviction evidence. Platform-controlled page-cache eviction and storage-cold remain explicitly unsupported future experiment types.

Local validation: 86 library tests, 8 CLI tests, Clippy with warnings denied, rustfmt, docs, AGENTS links, and action-pin guards all pass.

crates/omnigraph-bench/tests/cli.rs
aaltshuler: Fixed in `400594c5`. The CLI expectation is gated with `#[cfg(debug_assertions)]`, while cheap runner unit tests retain build-attestation coverage. A release-profile check of the crate also passes without entering the checked 800k-row suite.

crates/omnigraph-bench/src/environment.rs
aaltshuler: Fixed in `400594c5`. macOS host probing now uses native `statfs` for mount facts, and the remaining `diskutil` probe runs in a dedicated process group under a 10-second deadline with concurrent, bounded stdout/stderr capture. Timeout or monitor failure kills and reaps the direct child plus descendants. Regression tests cover a hanging descendant and independent output caps on both pipes.

crates/omnigraph-bench/src/environment.rs
aaltshuler: Fixed in `400594c5` by removing the `df` parser entirely. Native `statfs` now returns the filesystem, exact mount path, block size, and available-block count without whitespace tokenization. Tests cover repeated spaces in mount paths, malformed/non-UTF-8 fields, and byte-count overflow.

crates/omnigraph-bench/src/worker.rs
aaltshuler: Fixed in `400594c5`. Pre-measurement head capture now emits `protected_head_capture_failed`, which maps to `Prepare`; post-measurement `verification_failed` remains `Verify`. The owning worker test proves the pre-Ready failure retains its useful stage.

aaltshuler:
Addressed the complete review in `400594c5` (including the four inline threads):

- `CountingStore::get_ranges` now counts once and forwards the vectored call unchanged; the trait impl denies missing methods and a spy-store regression proves no scalar fallback.
- The cache vocabulary is honest: `process-cold` means a fresh process with no warm-up program, while OS/device/object-store cache state is explicitly uncontrolled and no page-cache-cold claim is made.
- Fixture build, full byte digest, APFS clone-freeze, and active-tree retirement run in a watchdog-bounded process group. Timeout and panic paths kill/reap descendants; malformed or forged handoffs fail closed, and the parent re-derives summary facts before acceptance.
- Runner-v1's full three-branch verification is documented as O(store), outside timing; receipt-based O(delta) certification remains with versioned-S3 reset.
- Missing child evidence now means containment is unproved and forces quarantine.
- SUT evidence now includes the exact executable SHA-256, source commit/dirty state, effective `LANCE_MEM_POOL_SIZE`, release/optimization/debug facts, and parent-observed peak RSS.
- Case admission and construction share one change-mix function; child elapsed time is bounded by the parent clock.
- macOS mount evidence uses native `statfs`; `diskutil` has deadline-driven process-group containment and bounded concurrent capture.
- Reset tests no longer require creating a Unix socket and clonefile tests verify/skip on non-APFS temp volumes.
- The new engine benchmark source guards pass. Diagnostics now use the correct lifecycle stages, shape terminology, YAML vocabulary, supported-history guidance, and expected/observed admission values; unreachable timer validation was removed.
- The synthetic fixture is now builder v2: an equal node/edge table mix with stable edge IDs, deterministic ring endpoints, edge-only divergence, endpoint-aware warming, and exact node+edge verification across source, target, and main.

Validation: 109 crate tests, 9 CLI tests, 7 engine source guards, all-target Clippy with warnings denied, rustfmt, release-profile check, docs/AGENTS checks, and workflow pin guards all pass locally.

PR #557, opened 2026-08-26
feat(bench): add durable benchmark telemetry

aaltshuler:
## Stack

Depends on #556. The base branch is intentional so reviewers see only the durable telemetry slice.

This replaces the durable-recording and read-model portions of closed PR #537 on top of the strict YAML foundation and identical-state runner.

## What changed

- Defines one strict canonical run-record-v1 contract with complete experiment, SUT, machine, backend, fixture, acquisition, raw-sample, phase, and logical-call evidence.
- Captures process-effective machine identity in every isolated worker and requires exact equality across repetitions.
- Publishes canonical JSON through a content-addressed object plus one immutable invocation pointer. Publication, verification, reconciliation, and inventory traversal are descriptor-rooted, bounded, nonblocking, no-follow, and crash-durable.
- Adds archive verify and candidate-specific archive reconcile commands. Indeterminate pointer durability returns the exact identity needed to close the ambiguity.
- Rebuilds a disposable OmniGraph projection from the JSON authority into immutable generations and atomically publishes CURRENT only after complete inventory and projected-row verification.
- Adds bounded, generation-pinned point and run queries; arbitrary GQ is not exposed.
- Streams archive-mode suite recording one completed run at a time so CLI memory does not grow with the suite.
- Strengthens frozen fixture proof: exact logical content, schema, index absence, lineage recipe, shared fork, physical digest, and exact post-merge verification.
- Records honest build evidence: exact executable digest and linked engine features, while explicitly marking effective LTO/codegen/strip settings unproved until a controlled AWS build receipt exists.

## Review shape

1. `feat(engine): expose benchmark evidence witnesses` is the small read-only engine seam for raw index-section and linked-feature evidence.
2. `feat(bench): add durable benchmark telemetry` owns the canonical record, archive, projection, CLI, evidence capture, docs, and tests.

## Correctness properties

- JSON archive remains the sole telemetry authority; the database can be deleted and rebuilt.
- A valid archived record is guaranteed projectable. Counter algebra and every projected sum are checked before publication, and an adversarial maximum-escaping record is exercised through the real 64 KiB projection-row guard.
- Root replacement, symlink, FIFO/special-file, immutable-path collision, partial publication, stale CURRENT, and non-UTF recovery paths fail closed with typed evidence.
- Human and JSON failure modes both preserve complete reconciliation and quarantined-workspace context.
- Logical store calls are never presented as physical requests, RTTs, or cloud cost.

## Verification

- `cargo test -p omnigraph-bench --locked` — 155 active library tests, 3 CLI unit tests, 12 command tests
- `cargo clippy -p omnigraph-bench --all-targets --locked -- -D warnings -W clippy::dbg_macro`
- focused engine raw-index and benchmark-scenario contract tests
- strict engine Clippy
- canonical workspace test graph with engine and cluster failpoints — green, including 148 active failpoint tests
- rustfmt, docs, AGENTS link, and workflow pin guards
- clean committed `cargo build --release --locked -p omnigraph-bench`
- release CLI smoke: suite validate/plan, archive verify, projection rebuild, bounded point query

## Deferred

AWS orchestration and controlled build receipts, S3 fixture reset/version cleanup, fixture caching, server-mode execution, cold page-cache proof, physical request attempts, A/A floors, comparison/diff reports, and performance claims remain later slices.

<h3>Greptile Summary</h3>

This PR adds durable benchmark telemetry with canonical archived records, crash-safe publication and reconciliation, rebuildable query projections, and stricter source, machine, fixture, and build evidence.
- Adds a content-addressed JSON archive and immutable invocation pointers.
- Adds verified projection rebuilds and bounded point/run queries.
- Streams suite recording and revalidates source provenance at publication.
- Exposes engine witnesses needed for benchmark evidence.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| crates/omnigraph-bench/build.rs | Captures bounded build and source-provenance evidence; runtime checkout revalidation prevents stale cached dirty-state evidence from entering durable records. |
| crates/omnigraph-bench/src/source_provenance.rs | Revalidates the compiled checkout’s commit, index, tracked bytes, and untracked-file inventory before recording. |
| crates/omnigraph-bench/src/archive.rs | Implements descriptor-rooted content-addressed archive publication, verification, reconciliation, and inventory traversal. |
| crates/omnigraph-bench/src/projection.rs | Rebuilds disposable immutable projection generations from archived JSON and verifies them before publishing CURRENT. |
| crates/omnigraph-bench/src/record.rs | Defines and validates the canonical durable benchmark run-record contract. |
| crates/omnigraph-bench/src/bin/omnigraph-bench.rs | Adds archive and projection commands while enforcing source revalidation at recording initialization and publication boundaries. |

</details>

```mermaid
flowchart LR
  Runner[Benchmark runner] --> Validate[Validate source, machine, fixture, and build evidence]
  Validate --> Record[Canonical run-record-v1 JSON]
  Record --> Object[Content-addressed archive object]
  Object --> Pointer[Immutable invocation pointer]
  Object --> Rebuild[Verified projection rebuild]
  Rebuild --> Generation[Immutable projection generation]
  Generation --> Current[Atomic CURRENT publication]
  Current --> Query[Bounded point and run queries]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(bench): harden durable telemetry"](https://github.com/modernrelay/omnigraph/commit/fcd60b4c2a4d2edea950b301fb89ebeb36fdc244) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57042072)</sub>


crates/omnigraph-bench/src/archive.rs
```diff
@@ -0,0 +1,3417 @@
...
+                    } else {
+                        remaining.set(count - 1);
+                        true
+                    }
+                })
+            {
+                last_message = "injected invocation-pointer directory fsync failure".to_string();
+                continue;
+            }
+            match self.sync_relative_parent_directory(relative) {
```
aaltshuler: **[P1] Do not accept retry success as proof of durable publication.** A substantive parent-directory `fsync` error leaves the visible link durability unknown; reopening the directory and observing a later success does not prove the failed update reached stable storage. The current transient test skips the first sync call, so it does not model writeback failure. Retry only a true interruption on the same descriptor; otherwise return the typed `possibly_published` outcome.

crates/omnigraph-bench/src/archive.rs
```diff
@@ -0,0 +1,3417 @@
...
+        )
+    })?;
+    let record = parse_canonical_record(&record_bytes).map_err(|error| {
+        ArchiveError::new(
+            "archive_record_invalid",
+            Some(&object_path),
+            error.to_string(),
+        )
+    })?;
+    let observed_sha256 = record_content_sha256(&record).map_err(|error| {
```
aaltshuler: **[P2] Hash the canonical bytes already proved here.** `parse_canonical_record` deep-validates and reserializes the record to prove equality with `record_bytes`; `record_content_sha256` repeats that validation and serialization, for records up to 64 MiB. Hash the accepted byte slice directly. The new-generation projection path also rereads authority twice, so consider a bounded staging spool rather than retaining or revalidating the full archive.

crates/omnigraph-bench/build.rs
```diff
@@ -1,6 +1,358 @@
...
+    for raw_line in manifest.lines() {
+        let line = raw_line.trim();
+        if line.starts_with('[') {
+            in_release = line == "[profile.release]";
+            continue;
+        }
+        if !in_release || line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+        let (key, value) = line.split_once('=')?;
```
aaltshuler: **[P2] Parse the manifest as TOML instead of scanning lines.** Any non-assignment inside the exact-spelled `[profile.release]` section aborts through this `?`, and valid TOML forms such as single-quoted values, dotted keys, or unrelated complex values degrade the build evidence to unknown and refuse otherwise valid runs. A typed TOML parse gives exact fail-closed behavior without accepting only the current formatting.

The rerun set also deserves tightening: it emits the repository `.git` path plus every tracked/untracked source path, currently 409 and permitted up to 100,000. A missing tracked path remains permanently changed to Cargo, while a primary checkout makes the broad `.git` directory sensitive to unrelated housekeeping. Keep explicit HEAD/index/ref watches and use the nearest existing parent when a tracked path is missing.

crates/omnigraph-bench/src/projection.rs
```diff
@@ -0,0 +1,4680 @@
...
+        )
+    })?;
+    let final_generation = generations_root.join(&inventory.generation_id);
+    enforce_generation_retention(
+        &retained_generations,
+        &inventory.generation_id,
+        &projection_root,
+    )?;
+
+    if final_generation.exists() {
```
aaltshuler: **[P2] Preserve metadata uncertainty instead of treating it as absence.** `Path::exists` converts EACCES, EIO, ELOOP, and broken-link errors to `false`, which can turn an indeterminate generation into a rebuild or mask the original rename failure. Use the reconciled inventory for the first check and typed `symlink_metadata` in the lost-rename arm, distinguishing only NotFound. That arm also verifies the same generation again at line 1152; retain and reuse the first verified manifest.

crates/omnigraph-bench/src/runner.rs
```diff
@@ -193,19 +220,108 @@ pub struct RunExecution {
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 pub struct BuildEvidence {
+    pub source_commit: String,
+    pub source_tree_dirty: bool,
     pub cargo_profile: String,
-    pub opt_level: String,
+    pub cargo_opt_level: String,
```
aaltshuler, line 265: **[P1] Bump `RUNNER_OUTPUT_VERSION` to 2 in this PR.** Relative to separately landing PR 556, this struct renames `opt_level` to `cargo_opt_level`, adds required build and machine fields, nests fixture evidence, and changes both success and failure envelopes, while the emitted version remains 1. Add a compact JSON-shape/version regression so future breaking changes cannot remain under the same version.

aaltshuler (commented):
Validated the telemetry-slice findings against this PR head. Blocking: directory-fsync retry cannot prove durable publication after a substantive error, and the breaking runner JSON rewrite must bump its version while PR 556 remains a separate landing unit. The remaining inline findings are confirmed P2/P3 robustness, efficiency, provenance, or test gaps. I did not duplicate the existing Greptile dirty-attestation thread.

crates/omnigraph/src/instrumentation.rs
```diff
@@ -33,6 +33,20 @@ use lance::io::WrappingObjectStore;
...
 use crate::storage::{ListDirBounds, StorageAdapter};
 
+/// Cargo features compiled into this exact `omnigraph-engine` artifact.
+///
+/// This read-only build seam lets benchmark and diagnostic binaries report
+/// dependency features from the crate that owns them. A dependent crate's
+/// `cfg(feature = ...)` namespace cannot observe features enabled directly on
+/// `omnigraph-engine` by Cargo's workspace feature graph.
+#[doc(hidden)]
+pub const fn enabled_engine_cargo_features() -> &'static [&'static str] {
```
aaltshuler: **[P2] Make feature attestation complete by construction.** The current list is accurate because `failpoints` is the only engine feature, but the runner/worker comparison is circular because both binaries call this same function. A future Cargo feature can be enabled yet omitted silently. Generate the sorted inventory from `CARGO_FEATURE_*`, or at minimum add a source guard against the engine Cargo.toml feature table.

crates/omnigraph-bench/src/bin/omnigraph-bench.rs
```diff
@@ -506,37 +988,258 @@ struct RunnerFailure<'a> {
...
+#[derive(Debug)]
+struct RecordingContext {
+    archive_root: PathBuf,
+    session_id: Ulid,
+    session_id_string: String,
+}
+
+impl RecordingContext {
+    fn new(archive_root: PathBuf) -> Result<Self, RecordingError> {
+        preflight_archive_publication(&archive_root).map_err(RecordingError::from_archive)?;
```
aaltshuler: **[P2] Run immutable eligibility checks before archive I/O.** This preflight creates and fsyncs the archive layout before source-commit, dirty-tree, release-build, and runtime-override refusals that can never succeed for the current binary. I reproduced `release_build_required` after archive shards had already been created. Split out a pure execution-precondition check and touch the archive only after it passes.

azimafroozeh:
Excellent!
**[P1]** The PR throws away measurements that matter. A run that completes 17 of 20 verified repetitions produces no durable record: record.rs:1029 forces observed == requested, so all 17 samples die with the failure JSON. Those are exactly the samples we can least afford to lose, long cells cost hours to acquire and rule 3 needs 20 of them before a p95 may even be stated. Fix: let a record hold fewer repetitions than requested, observed <= requested, with typed evidence of why the run stopped early; each record stays immutable from birth, rows never append. The RFC's "repetition rows append" sentence gets amended to match in the same commit as the other pending edits.

**[P1]** The record can't distinguish engines whose configuration differs by environment. The RFC's SUT definition names "engine configuration" as SUT identity, but LANCE_MEM_POOL_SIZE (consumed by lance-datafusion, set to 1GiB by .cargo/config.toml for exactly the cargo-run invocation the harness recommends) is invisible: a 1GiB-pool record and a 150MB-pool record are byte-identical. Fix: store engine-consumed configuration env by name and value in EngineConfigurationV1 (the LANCE_* and OMNIGRAPH_* namespaces, sorted), keep ignoring everything else since env can carry credentials and the archive is permanent, and qualify the v0.10.0.md "refusing unmodeled runtime overrides" sentence to match.

aaltshuler:
Updated in `fcd60b4c` and resolved the eight inline review threads.

Key fixes:
- source provenance now verifies raw tracked bytes, ignored/untracked inputs, index flags, repository/worktree identity, replacement refs, and the complete engine feature registry;
- fixture and repetition children run in protocol-owned scratch directories with a closed environment, and executable/protocol inputs reject symlinks and special files without blocking;
- archive publication now closes ancestor-directory durability, ambiguous-publication reconciliation, immutable inventory, and lazy per-record healing;
- durable records distinguish complete from censored acquisition, require supervisor peak RSS, separate claim eligibility, and bind typed projection metadata;
- fixtures are real node-plus-edge graphs with deterministic logical evidence and exact content verification.

Validation on the pushed tree:
- `cargo test --workspace --locked --features omnigraph-engine/failpoints,omnigraph-cluster/failpoints`
- both workspace Clippy graphs with `-D warnings -W clippy::dbg_macro`
- `cargo test -p omnigraph-bench --locked`
- all repository documentation and workflow guard scripts
- clean-commit release build plus case-validation and suite-plan CLI smoke tests

Two independent adversarial audits found no remaining concrete blockers. CI is now rerunning against the updated head.

Issue #563, opened 2026-08-28
performance: bm25/rrf ranked read + a join materializes the joined column corpus-wide → Arrow 2 GB offset overflow at ~700k+ rows


FlorianSeemann:
## Summary

A ranked read (`bm25(...)` or `rrf(...)`) that also traverses/joins to another node type makes the planner materialize the **joined variable-width column for the whole corpus** before it prunes to the top-`limit`. Once that column exceeds Arrow's 2 GB per-column offset limit (i32 offsets on a `Utf8`/binary array), the query fails with an `Offset overflow`-class error. In practice this kicks in on a large text corpus (~700k+ rows) and forces dropping the join from otherwise-natural recall queries.

## Environment

- Engine 0.8.1 (downstream build; the relevant planner/scan paths are unmodified from upstream).
- Single-node `omnigraph-server`, S3-compatible storage.
- Corpus: a `Chunk` node type, ~1.05M rows, `text: String @index` (BM25) + `embedding: Vector(3072) @index` (ANN), with a `ChunkOfArtifact: Chunk -> Artifact` edge.

## Repro (shape)

```gq
// Fails at corpus scale — the rrf/bm25 leg + the join materialize Chunk.text corpus-wide:
query recall($q: String) {
    match {
        $c: Chunk
        $c chunkOfArtifact $a          // <-- the join is what tips it over
        search($c.text, $q)
    }
    return { $c.slug, $c.text, $a.slug, $a.name }
    order { rrf(bm25($c.text, $q), nearest($c.embedding, $q)) }
    limit 20
}
```

The same query **without** the `chunkOfArtifact` join (rank the chunks alone, recover the artifact from the chunk slug afterward) runs fine — because a `nearest()`-only path prunes to top-k first, but the bm25/rrf leg + join does not.

## Impact

We had to drop the `chunkOfArtifact` join from our keyword and hybrid recall queries entirely and reconstruct artifact identity from the chunk slug downstream. That's a real loss of expressiveness on the hot retrieval path — you can't rank-and-join in one query at corpus scale.

## Suggested direction

Push the rank + `limit` down so the join only materializes the top-k rows (rank first on the base table, then join the survivors), rather than materializing the joined column across the full corpus. If 0.9's filter/scan-pushdown work already changes this, it'd be great to confirm whether ranked-read-with-join is now bounded — we haven't yet re-tested on 0.9.

FlorianSeemann:
[...]

PR #546, opened 2026-08-23
rfc: add RFC 0040, the system column namespace

azimafroozeh:
## What & why

This PR adds RFC 0040, which frees the property names `id`, `src`, and `dst` for user schemas and moves OmniGraph's implicit stored columns into a reserved namespace. Motivated by #529: a ported schema whose natural key is named `id` collides with the implicit physical id column, failing with a misleading error when declared with `@key` and silently producing a duplicate physical column without it.

- Reserve the leading-underscore property namespace for the system; the per-name Lance reservations collapse into one prefix rule, and future system columns need no further reservations or versions.
- Newly created graphs spell the system columns `__id`/`__src`/`__dst`; every existing graph keeps its current spellings through per-graph resolution keyed on its schema IR version, so the design contains no migration.
- The query language gains meta-fields (`$p.@id`, `$e.@src`, `$e.@dst`) that work on every graph; `$p.id` keeps its exact current behavior on existing graphs (with a deprecation lint) and refers only to a user property on new ones.
- API payloads carry each graph's own column names; an explicit per-graph upgrade renames the columns and mechanically rewrites the graph's stored queries in one schema apply.

## Backing issue / RFC

- [x] This PR is the RFC (`docs/rfcs/0040-system-column-namespace.md`); motivated by #529. A companion bug-fix patch reserving the three names lands separately before the implementation: it is the fence that makes the RFC's coexistence guarantees provable.

## Checklist

- [x] Change is focused (adds one RFC document, no source changes)
- [x] Tests added/updated for behavior changes (none: docs-only, no behavior change)
- [x] Public docs updated if user-facing surface changed (the RFC is the added document; user docs change with the implementation)
- [x] Reviewed against docs/dev/invariants.md: no Hard Invariant weakened; the RFC's Invariants section names the one knowingly brushed deny-list item (on-disk format change) and carries its refusal and compatibility evidence plan

## Local verification

- no build or test commands: docs-only RFC addition, no source touched
- vocabulary guard not run: the diff touches only `docs/rfcs/`, outside the guarded surfaces

## Notes for reviewers

- No flag day: at the release boundary, existing graphs, queries, and clients see no behavior change beyond a deprecation lint; each graph adopts the new spellings at its own explicit upgrade.
- The new binary permanently accepts exactly two schema IR vintages; old binaries refuse new-vintage graphs through the existing strict version validation.
[...]

PR #584, opened 2026-08-31
docs(rfc): add RFC 0043, GQ logic tests

azimafroozeh:
## What & why

RFC 0043 makes query-behavior tests one self-contained text file per case (`.gqt`): a header (issue anchor, `red_on:` provenance), a `.pg` schema, JSONL seed rows, then steps (`--- query` / `--- mutate` / `--- restart`, loopable), each with params and an expected outcome (rows, affected counts, or an error substring). One test target, `crates/omnigraph/tests/gq_logic_tests.rs`, runs each case against a fresh temporary store through public engine surfaces only; no second toolchain. An enforcement ladder makes the medium stick: AGENTS.md contract sentences, a CI gate holding issue-closing fix PRs to a matching regression, and a `no-repro` waiver label.

Why now: review of the #563 fix surfaced a pure input-to-output defect (aggregates computed over the capped BM25 scan window) a twenty-row logic test case would have caught before review. The format adopts the sqllogictest lineage (SQLite, DuckDB, DataFusion) at its mature state and omits its documented mistakes (result hashing, type strings, bare any-error expectations).

## Backing issue / RFC

- [x] Is an RFC PR: `docs/rfcs/0043-gq-logic-tests.md`

## Checklist

- [x] Change is focused (one RFC plus its registry row)
- [x] Tests added/updated for behavior changes (N/A: proposal only)
- [x] Public docs updated if user-facing surface changed (the RFC and its registry row are the docs; no engine surface changes)
- [x] Reviewed against [docs/dev/invariants.md](../blob/main/docs/dev/invariants.md) — no Hard Invariant weakened, no deny-list item hit

## Local verification

- `python3 scripts/check-docs.py` — green (front matter, registry row, links)
- `cargo test --workspace` — not run: docs-only diff

## Notes for reviewers

- Number 0043: open PRs hold 0040 (#546) and 0042 (#552); next-available moves to 0044. If #552 merges first, `docs/rfcs/README.md` conflicts on the index table; keep both new rows in number order, 0044 stands.
- v1 takes the full corpus role (feature cases, mutate and restart steps, loops), argued in Alternatives. Run-twice traversal-mode verification is deferred past #544 (Compatibility); `proptest_equivalence.rs` keeps owning equivalence.
- Rollout after acceptance: two implementation PRs — harness plus first corpus, then the enforcement gate.

<h3>Greptile Summary</h3>

This documentation-only PR proposes RFC 0045, defining a self-contained `.gqt` format and an enforcement model for query-behavior regression tests.
- Specifies schema, seed, query, mutation, restart, loop, comparison, and bless semantics.
- Defines CI execution, fix-regression gating, waiver handling, and heavy-reproduction tiers.
- Registers RFC 0045 in the RFC index.

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| docs/rfcs/0045-gq-logic-tests.md | Adds the RFC describing the `.gqt` format, runner semantics, CI enforcement, compatibility boundaries, evidence plan, and rollout. |
| docs/rfcs/README.md | Adds the RFC 0045 registry entry with frontmatter-consistent metadata. |

</details>

```mermaid
flowchart LR
  Case[.gqt case] --> Parse[Fail-closed parser]
  Parse --> Init[Fresh temporary store]
  Init --> Seed[Load JSONL seed]
  Seed --> Index[Ensure required indexes]
  Index --> Steps[Run query / mutate / restart steps]
  Steps --> Compare[Normalize and compare outcomes]
  Compare --> Result[Report case result]
  PR[Issue-closing PR] --> Gate[Fix Regression Gate]
  Gate -->|matching regression| CI[GQ Logic Tests]
  Gate -->|no-repro label| Waiver[Maintainer waiver]
  CI --> Result
```
</details>

<sub>Reviews (5): Last reviewed commit: ["docs(rfc): address review: gate guarante..."](https://github.com/modernrelay/omnigraph/commit/a911c8c15d3b50d339810f45cb8d129b97a6c468) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58652103)</sub>

### DIFF
--- a/docs/rfcs/0045-gq-logic-tests.md
+++ b/docs/rfcs/0045-gq-logic-tests.md
@@ -1,0 +1,700 @@
+---
+rfc: "0045"
+title: "GQ logic tests"
+track: maintainer
+status: draft
+implementation: not-started
+authors:
+  - azimafroozeh
+created: 2026-08-29
+updated: 2026-08-30
+discussion: null
+supersedes: []
+superseded_by: []
+blocked_on: []
+---
+
+# RFC 0045: GQ logic tests
+
+> A term set in ***bold italics*** is being defined at that exact spot; it is
+> used plain everywhere after.
+
+## Summary
+
+Query-behavior tests become small text files, and the logic test corpus
+is their default home: regressions for merged bug fixes, and feature
+cases (cases with no issue anchor) for new or existing behavior alike. A
+***logic test case*** is a single self-contained file (extension `.gqt`)
+holding a `.pg` schema, seed rows as JSONL, and one or more ***steps***:
+a read query or a mutation, each with its params and its expected outcome
+(rows, affected counts, or an error message); a case may also restart the
+store between steps and repeat a step group over a value list. One test
+target, `crates/omnigraph/tests/gq_logic_tests.rs`, walks
+`tests/gq_logic_tests/*.gqt` and runs each case against a fresh temporary
+store: init, load, index, then the steps in order.
+
+Around the harness sits an enforcement ladder: AGENTS.md contract
+sentences making logic tests the default medium and holding every issue
+fix to a regression, a CI check holding issue-closing fix PRs to a
+matching regression, and a `no-repro` waiver label for fixes that cannot
+carry one.
+
+The boundary that does not change: Rust tests keep owning mechanism
+assertions, scale symptoms, and cases needing process environment
+(examples in the AGENTS.md sentences below). No second toolchain enters
+the repo, and the harness calls only public engine surfaces.
+
+## Motivation
+
+Query-behavior tests today are hand-written Rust (the `_issue_NNN`
+convention, for example `mutate_expected_head_precondition_issue_365` in
+`tests/writes.rs`): every regression pays Rust-test cost, enforcement
+stays review-only, and nothing structural forces a fix PR to carry a
+regression. The concrete evidence is #563: review of its fix surfaced a
+further pure input-to-output defect, aggregate returns computed over the
+capped BM25 scan window, so `count($c)` reported the window size, not the
+match count. A logic test case (twenty seeded chunks, one aggregate
+query, expected `total: 20`) expresses it exactly and would have caught
+it before review.
+
+Cheap tests get written; this RFC adopts the precedent's mature
+state. The sqllogictest
+lineage (SQLite's format, extended by DuckDB and DataFusion) became the
+primary test medium: DuckDB routes essentially all query-behavior tests
+through data files ("We strongly prefer tests to be written using the
+sqllogictest framework") and reserves C++ tests for what the format
+cannot express. The lineage's load-bearing features (statements between
+queries, loops, restart records) are v1 features here; its documented
+mistakes (result hashing, type strings, bare any-error expectations) are
+omitted; its run-both-plans verification is deferred (Compatibility below).
+`docs/dev/testing.md` rule 2 already points the same way, and the corpus
+(small schemas, seeds, queries with known-correct answers) doubles as
+seed material for the DST generators.
+
+## User and operational behavior
+
+Authoring a regression for a fixed issue:
+
+1. Write `tests/gq_logic_tests/issue_NNN_short_name.gqt`.
+2. Run it on the unfixed build and watch it fail; record what failed in the
+   `# red_on:` header line, mandatory for issue-anchored cases (a logic
+   test nobody watched fail guards nothing).
+3. Fix the bug; the case goes green.
+
+Authoring a ***feature case*** (a case written for new or existing
+behavior with no failure to witness): write
+`<short_name>.gqt` with `# issue: none`; `red_on:` is omitted, or kept when
+the case did witness a red state during development.
+
+Running:
+
+```bash
+cargo test -p omnigraph-engine --test gq_logic_tests
+```
+
+The target prints one line per case (`ok issue_563_aggregate_uncapped` /
+`FAIL issue_563_underfill_retry`) and fails at the end with the list of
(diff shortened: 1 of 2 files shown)


docs/rfcs/0045-gq-logic-tests.md
```diff
@@ -0,0 +1,706 @@
...
+executable addition in the diff under `crates/*/tests/**`, checked
+independently per issue: an added `.gqt` case in the logic test corpus
+whose file name carries `issue_N`, an added `# issue: N` header line in a
+corpus case, or an added Rust line defining a function whose name carries
+`issue_N`, where `N` must be followed by a non-digit or by the end of the
+line or path (`issue_5630` never matches issue 563; `issue_563_underfill`
+does). A comment, string, or fixture line mentioning the issue never
+satisfies the gate. Each accepted shape executes: the corpus walker runs
+every case and refuses a malformed one, and the workspace clippy gate
+refuses a dead test function; whether the test asserts the right thing
```
aaltshuler: RFC-level concern: this paragraph promises that every accepted Rust regression executes, but the proposed required PR job only runs `gq_logic_tests`, and Clippy only compiles test targets. A function line can satisfy the diff check without being a registered or executed test. The `crates/*/tests/**` restriction also conflicts with `docs/dev/testing.md`, which assigns several packages' regressions to in-source tests. Please define the gate in terms of a registered and executed regression at the declared owner boundary, or narrow the guarantee accordingly; the implementation details can remain open.

docs/rfcs/0045-gq-logic-tests.md
```diff
@@ -0,0 +1,706 @@
...
+  the comparison is positional, which makes it multiset equality:
+  duplicate rows are legal and compare by multiplicity. GQ guarantees no
+  order without an `order` clause, and ordering assertions are a
+  documented flakiness source in comparable suites.
+- `expect ordered`: positional comparison, allowed only when the query has an
+  `order` clause; the harness checks the parsed declaration and refuses
+  `ordered` without one. With an explicit `order` the engine's output order is
+  a total, deterministic, shipped contract (`apply_ordering` appends the bound
+  entities' key columns as an ascending tie-break), so positional comparison
+  is stable.
```
aaltshuler: RFC-level semantic issue: `expect ordered` is allowed for any `order` clause because this claims `apply_ordering` supplies a total key tie-break. That is not true for ranked search: nearest and BM25 skip `apply_ordering`, and RRF sorts by score alone. Equal distances or scores therefore have no stable order. Please choose the contract here—either ranked queries gain a deterministic tie-break, or the format disallows ordered ranked expectations—without needing to specify the implementation.

docs/rfcs/0045-gq-logic-tests.md
```diff
@@ -0,0 +1,706 @@
...
+Ranking scores stay unprojected in logic tests (existing search-test
+practice): assert the resulting row order, never the score values.
+
+### Runner mechanics
+
+The walker is a single `#[tokio::test(flavor = "multi_thread")]` entry
+point (tokio is already every engine integration test's runtime) that
+lists `tests/gq_logic_tests/*.gqt` rooted at `CARGO_MANIFEST_DIR`
+(the `forbidden_apis.rs` walk precedent) and spawns each case as a task
+into a `tokio::task::JoinSet`. Case concurrency comes from that task set,
```
aaltshuler, line 539: RFC-level invariant issue: this chooses to spawn the entire corpus into one `JoinSet`. Because each case creates a store and may build indexes, concurrency grows with corpus size while libtest sees only one test. The RFC only needs to state that case execution is bounded; the actual worker count, limits, and timeout values can remain implementation details.

aaltshuler (commented):
Three RFC-level comments: the regression-gate guarantee, ranked ordering semantics, and bounded runner concurrency.

docs/rfcs/0045-gq-logic-tests.md
azimafroozeh: Agreed, with one correction: non-aggregate nearest/bm25 do get the .id tie-break (search_score_orderings feeds apply_ordering). The unstable ones are rrf() (execute_rrf_fusion sorts by fused score alone) and search-led orders with aggregates (execute_query_once skips both sort branches).

Went with the format side. ordered now requires a deterministic row set plus a total order on it; the harness refuses it on those two shapes. Distinct scores at the cutoff are the author's job, a tie there changes the set.

docs/rfcs/0045-gq-logic-tests.md
azimafroozeh: Agreed. The RFC now says only that the number of cases running at once is capped, independent of corpus size, and that each case has a time budget; the numbers are implementation details. Two ideas behind the budget: cases should be small and fast, well under a second, and that is what every PR runs; a case that genuinely needs to be slow goes to nightly instead. Each case reports its elapsed time so the slow ones are visible before they fail the budget.

docs/rfcs/0045-gq-logic-tests.md
azimafroozeh: Taking the narrow option. The gate stays a diff check, and the paragraph now says what that guarantees: a .gqt case is guaranteed to run, since every file in the corpus runs on every PR; a Rust test is only checked by name, and whether it runs and asserts the right thing stays with review. The idea is that .gqt is the default and the Rust path is the exception that needs a reason. The gate also accepts regressions wherever testing.md says the package keeps them, in-source included. A gate that proves execution by running the named tests is recorded in Alternatives as the upgrade if a named test ever turns out not to run; not taken now because the gate re-checks on every PR-body edit, and that should stay cheap.

PR #591, opened 2026-09-01
Bound nearest ANN probe budget

### LINKED ISSUE #567: performance: nearest() sets no ANN probe budget — Lance's unbounded default probes every IVF partition (289s vs 7.5s at 852k × 3072-dim)
## Summary

`nearest()` sets no probe budget on the Lance scanner. Lance's Rust scanner defaults to `minimum_nprobes = 1` with **no maximum**, and on the rev we measured (9.0.0-beta.15, the 0.8.1 pin) that degenerates into probing **every IVF partition** — i.e. a full read of the raw-vector index payload per query. On a large vector column over object storage this turns indexed ANN into something close to a flat scan.

Measured on our production corpus (852k rows × `Vector(3072)`, S3-compatible storage, single-node `omnigraph-server`): **289 s per `nearest()` query unbounded vs 7.5 s with a pinned probe budget** — a ~38× difference, on the same index.

For contrast, pylance has shipped `nprobes = 20` as its default for years; the Rust scanner path diverges from that.

## Where

`crates/omnigraph/src/exec/query.rs`, the `search_mode.nearest` branch of the node-scan configure closure: `scanner.nearest(prop, &query_arr, k)` is called with no follow-up probe configuration. Still the case on `v0.9.0` and current `main` (no `nprobes` handling anywhere in the exec path). We haven't re-measured on Lance 10 — if the scanner defaults changed there this may be moot, but the engine still pins nothing, so the behavior is at Lance's mercy either way.

## Downstream fix (works, in production since 2026-08-04)

Commit `e115c09` on our 0.8.1 build (`FlorianSeemann/omnigraph`, branch `p9/v0.8.1-optimize-table-scope`):

- after `scanner.nearest(...)` (it no-ops with a warning before), pin the probe budget to the pylance-parity default;
- operator-tunable via an env var (`OMNIGRAPH_ANN_NPROBES`) so recall/latency can be tuned without a rebuild.

Happy to adapt it into a PR against `main` — a config/cluster-level knob instead of an env var is fine by us, the important part is that the default is bounded.

## Related (can split out if wanted)

Same family, same downstream branch (commit `df7f108`): sizing IVF partitions from the table's row count at build time, plus `optimize --retrain-vector-index` to re-train a partitioning that has drifted badly from the table it serves. At our growth rate the fixed default partitioning drifts within weeks; retraining restored latency without a full rebuild. Say the word and I'll file it separately with numbers.

mikemikimike:
## Summary

- Keep Lance's adaptive ANN minimum at 1 and cap only `maximum_nprobes` at 20 by default.
- Preserve `OMNIGRAPH_ANN_NPROBES` as the deployment tuning knob.
- Retry detectably short standalone nearest scans and individual RRF nearest arms once without the cap.
- Add a deterministic Lance 11 reproduction that distinguishes centroid ranking from payload-partition search.

## Why

Lance 11 starts with one IVF probe, but its `early_pruning` heuristic can promote that minimum to every partition for off-centroid or near-equidistant queries. Calling `nprobes(20)` avoids the runaway expansion, but it also forces easy queries that need only one partition to scan 20.

The corrected implementation therefore sets only `maximum_nprobes(20)`. Easy queries retain Lance's one-partition fast path, while pathological adaptive expansion is capped. If the capped pass visibly under-fills the requested result count, OmniGraph retries once without the cap so standalone nearest queries and RRF vector arms can recover candidates.

## Reproduction

Using a deterministic 10,000-row, 32-dimensional IVF_FLAT index with 100 partitions:

- Centroid-aligned query: Lance default and max-only both load 1/100 payload partitions; fixed min=max 20 loads 20/100.
- Near-equidistant query: Lance default loads 100/100 payload partitions; max-only 20 loads 20/100.
- Selective filtered query: the bounded pass returns 20/50 rows; the uncapped retry returns 50/50.
- Large-`k` query: the bounded pass returns 2,000/2,001 rows, exercising the same generic under-fill retry.

The reproduction also records payload-partition loads, bytes, I/O operations, and requests rather than inferring work from centroid ranking alone.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p omnigraph-engine --lib --tests --benches --locked -- -D warnings`
- `cargo test -p omnigraph-engine --locked`
- `python3 scripts/check-fix-regression.py --self-test`
- Fix Regression Gate for issue #567 against `origin/main...HEAD`
- `cargo bench -p omnigraph-engine --bench scenarios -- --scenario ann-probe-budget --rows 10000 --dims 32 --selectivity 0.01 --k 50 --ann-partitions 100 --ann-probes 20`

Closes #567


mikemikimike:
Thanks for catching this. Fixed in e5e1a60 (https://github.com/mikemikimike/omnigraph/commit/e5e1a60ae1075df2c5ac8f4caae8c6a80bd94db3).

The ANN scanner now sets Lance minimum_nprobes and maximum_nprobes explicitly. A filtered nearest query keeps the bounded first pass, then retries once with the same minimum and maximum_nprobes unset when the result under-fills the requested limit. This preserves the existing latency/recall knob for normal queries while allowing late search to reach all partitions when prefiltering starves the capped pass. The effective min/max budget and retry are recorded by the query instrumentation seam.

Added the optimized multi-partition IVF behavioral regression, Lance plan guards for bounded versus minimum-only probing, and search documentation/release notes. Local validation after merging upstream/main: cargo +1.97.0 test -p omnigraph-engine --test search --locked (40 passed); cargo +1.97.0 check -p omnigraph-engine --locked (passed); cargo +1.97.0 clippy -p omnigraph-engine --lib --locked -- -D warnings (passed); cargo +1.97.0 fmt --all -- --check (passed). The full surface guard suite remains 27/33 because six pre-existing Windows/Lance local _refs/branches permission or e_tag guards fail in this environment.

ragnorc:
Closing this because it is no longer a valid fix for the current stack. Issue #567 was measured on Lance 9 beta, while OmniGraph now uses Lance 11's adaptive IVF search. Lance may rank every centroid, but it does not load and search every payload partition.

In a deterministic Lance 11 reproduction, the default searched 1/100 partitions for a normal k=50 query. Setting nprobes=20 searched 20/100 and caused roughly 7x more index I/O. With a selective prefilter, Lance adaptively searched 50 partitions and returned all 50 requested rows; the proposed cap returned only 20, then the retry searched 70 partitions in total. A large-k query returned 2,001 rows by default but only 2,000 with the cap.

The RRF path also bypasses the standalone retry, so the cap can silently omit ranking candidates. The original S3 performance problem may still need investigation, but it should first be reproduced on Lance 11 with the actual query and index shape.

ragnorc:
Correction to my earlier review: the first Lance 11 fixture queried exactly at a centroid, so its zero nearest-centroid distance masked the `early_pruning` expansion. A near-equidistant query reproduces the reported behavior: the default plan loads all 100 IVF payload partitions, while a max-only budget of 20 loads 20.

I have pushed the corrected implementation in d0a9979 and the deterministic reproduction/benchmark in 560b8c0. The implementation leaves Lance's adaptive minimum at 1, caps only the maximum, and retries detectably short standalone nearest scans and individual RRF vector arms without the cap. That preserves the one-partition fast path for easy queries while bounding the pathological case and recovering under-filled results.

The branch is current with `main`; formatting, clippy with warnings denied, the full `omnigraph-engine` test suite, the issue #567 regression gate, and the benchmark all pass locally.

Issue #583, opened 2026-08-31
bug: merge duplicates an edge added on both sides of the fork, and plain traversal hides the extra row


azimafroozeh:
### What happened

Fork a branch, insert the same edge on both sides (for example Knows Diana -> Alice on the branch and on main), merge back. The merge succeeds and the store now holds TWO physical rows for one logical relationship.

Mechanism, verified in source: the three-way streaming merge keys rows on the primary key id (stage_streaming_table_merge, crates/omnigraph/src/exec/merge.rs:2372 at ac620ee). Each insert mints a fresh ULID, so the two rows never meet in the cursor walk; each side reads as "born on one side only, take it", and MergeConflictKind::DivergentInsert (merge.rs:2700) only fires when the SAME id diverges. Nodes are protected: a Person inserted on both sides converges on its `@key`. Edges have no logical key, so the duplicate lands silently.

Why nobody sees it: plain traversal (match { $a: Person  $a knows $b }) deduplicates in its visited gate and returns the pair once. The bound-edge pattern ($a $e:knows $b), counts, aggregates, and exports all see 2 rows. Nothing errors; results are silently wrong after any convergent-edit merge.

Expected: a convergent edge merges to one row, or the merge surfaces a conflict. Two stored rows for one logical relationship should not be reachable.

### Steps to reproduce

Public API only. Reproduces identically on the local file:// backend and the in-memory object-store backend. The test uses the existing tests/helpers module and runs on local FS.

1. Save the test below as crates/omnigraph/tests/merge_duplicate_edge_repro.rs.

```rust
mod helpers;

use omnigraph::db::Omnigraph;
use omnigraph::loader::{LoadMode, load_jsonl};
use helpers::*;

const TEST_SCHEMA: &str = "node Person {\n    name: String @key\n    age: I32?\n}\n\nedge Knows: Person -> Person\n";

const TEST_DATA: &str = r#"{"type": "Person", "data": {"name": "Alice", "age": 30}}
{"type": "Person", "data": {"name": "Bob", "age": 25}}
{"type": "Person", "data": {"name": "Charlie", "age": 35}}
{"type": "Person", "data": {"name": "Diana", "age": 28}}
{"edge": "Knows", "from": "Alice", "to": "Bob"}
{"edge": "Knows", "from": "Alice", "to": "Charlie"}
{"edge": "Knows", "from": "Bob", "to": "Diana"}
"#;

const QUERIES: &str = r#"
[...]

PR #592, opened 2026-09-01
rfc: add RFC 0044, derived edge identity via @key on edge types

azimafroozeh:
## What & why

Adds RFC 0044, "Edge keys: derived edge identity" (maintainer track, draft). This is the design for issue #583.

What it is: an edge key is declared identity for a relationship. `@key(src, dst)` on an edge type says there is at most one such edge between two endpoints, and the endpoint pair is its identity. Nodes already have exactly this (`@key(name)` means one Person row per name); edges don't, the parser refuses `@key` on them.

Why we need it: without declared identity every edge insert mints a random ULID id, and branch merge keys rows on `id`. When two branches insert the same relationship (Alice knows Bob, once per side of a fork), the merge sees two different ids and keeps both rows (#583). The duplicate is silent: edge counts and bound-edge traversals go wrong and no conflict is raised. Branch-and-merge is the core write workflow, so any graph with two writers hits this.

The proposal: for a keyed edge type the id is derived from the key instead of minted, using the canonical encoding keyed nodes already use. Both branches then derive the same id and the merge converges to one row; the same key with different non-key properties surfaces a typed `divergent_insert` conflict instead of silence. Unkeyed edge types are untouched: parallel edges stay legal and existing graphs are unaffected.

Diff: the RFC file plus its registry row, no code.

## Backing issue / RFC

- [x] Is an RFC PR: docs/rfcs/0044-edge-keys.md, discussion on accepted issue #583

## Checklist

- [x] Change is focused (one RFC and its registry row)
- [x] Tests added/updated for behavior changes (N/A: docs only; the RFC's Evidence section names the test owners the implementation extends)
- [x] Public docs updated if user-facing surface changed (N/A here: the user-doc updates ride the implementation PR per the RFC's rollout plan)
- [x] Reviewed against docs/dev/invariants.md (the RFC's Invariants section maps the design onto invariants 6, 8, and 13 and argues the deny-list explicitly)

## Local verification

- `python3 scripts/check-docs.py`: OK, 114 files (frontmatter, registry row, and links validate)
- `cargo test`: not run, docs-only diff

## Notes for reviewers

- The `ir_version` question from review is settled for this RFC: the edge-key number is 3, fixed at acceptance, not at implementation time (Design change 6, Compatibility). The cross-feature scheme raised on #546 (independent features against one linear scalar) is deliberately out of scope here and deferred to a dedicated versioning RFC (tracked by RFC 0040's unresolved question 1).
- The registry's next available number advances to 0045.
- Frontmatter says `implementation: in-progress`: the full implementation (engine changes, tests, DST updates, user docs) is open as #593; #583 is closed there, not by this PR.


ragnorc:
Small fixes:
1. Title/body still say "RFC 0043" — that number is taken on main by full-text compatibility; the diff itself is already correctly `0044-edge-keys.md`. Retitle and fix the stale "next available advances to 0044" line in the body (the diff says 0045).
2. The shared `ir_version` scalar with #546/#548 deserves settling at acceptance rather than implementation — see the comment on #546 for the combination case a linear version can't express.
3. Rebase race with #584 on the registry's next-available line — whoever lands second rebases.

PR #725, opened 2026-09-16
RFC 0067 detached table commits, RFC 0068 graph commit record, and their Lance fences

ragnorc:
## Summary

- **RFC 0067, detached table commits**: every content effect stages as a Lance detached commit from its manifest pin, the manifest publishes `(target, staged, uuid)` pins in one CAS as today, and promotion replays the recorded transaction linearly. No recovery sidecar, no restore compensation, no HEAD-equals-pin invariant. Evidence in the RFC: 14 Lance probes (7 rerun on Lance 12.0.0-rc.1), an engine prototype for mutation, load, merge, ensure-indices and Optimize, S3 runs on RustFS, process kills, two-process races, an old-binary check, manifest-byte measurements, and a 312-cell failure-window matrix.
- **RFC 0068, graph commit record**: the sequel that replaces the `__manifest` fold with one immutable record per graph commit; includes the change-set idea for the feed and merge.
- **Fences** in `lance_surface_guards.rs` for the four Lance behaviours RFC 0067 depends on, including the self-conflict rule that decides which transaction kinds may be staged detached (Update, Delete, Rewrite, CreateIndex conflict with their twin; Append, ReserveFragments and delete-only UpdateConfig rebase).

No engine behaviour changes. The prototype is on `spike/rfc-0066-detached-prototype` for reference and is not for merge.

## Checks

- `cargo test -p omnigraph-engine --test lance_surface_guards` (42 passed, 1 ignored)
- `cargo clippy -p omnigraph-engine --all-targets -- -D warnings -W clippy::dbg_macro`
- `cargo fmt --all --check`, `scripts/check-docs.py`, `typos`, `scripts/check-agents-md.sh`, `scripts/check-workflow-action-pins.py`


ragnorc:

PR #728, opened 2026-09-16
storage: stamp v10 for RFC 0067 detached table pins, with the explicit route from v8 and v9

ragnorc:
RFC 0067 rollout step 2, the storage stamp. Stacked on #727 (base branch; retargets to `main` when the stack below it merges). No detached writer exists yet; this PR puts the fence in place so the first one (mutation and load, next PR) lands behind it.

## What changes

- **Stamp v10 is the only served format.** `INTERNAL_MANIFEST_SCHEMA_VERSION` and `MIN_SUPPORTED_INTERNAL_SCHEMA_VERSION` are both 10. v10 means a table registration may carry `omnigraph.staged_version` and `omnigraph.transaction_uuid` (RFC 0067); an older binary would open the absent linear target of a pending pin and misdiagnose it as reclaimed history, so the stamp refuses it before any open. Both RFC 0040 system column vintages are born at v10; the vintage is read from the schema IR, never from the stamp.
- **One explicit route from every 0.11.x graph.** `omnigraph upgrade` defaults to v10: v6 → v7 → v8 → v10, v7 → v8 → v10, v8 → v10 and v9 → v10. The new handler `detached-pins-v8-v9-to-v10` (intent protocol 3) restamps main and every live branch under one intent and touches nothing else; retired ancestors keep their stamp; a v8 or v9 source skips the retained-history qualification the v6/v7 handlers need. A receipt from an earlier step of the same route is recognised as a completed predecessor instead of a foreign attempt. `--to-format 7` and `8` remain for older executables; v9 is not a target.
- **The system-column respelling leaves the storage route.** It runs on a served graph (`omnigraph schema upgrade-system-columns`) with no stamp change: the stamp advance inside its intent is a no-op (`publish_stamp_advance` accepts `from == to`), its preflight no longer refuses branched graphs on the route, and the dead per-vintage stamp floor (`stamp_covers_system_columns`) is removed with the open-time re-read that fed it.
- **Refusal text.** A v6 to v9 stamp now names `omnigraph upgrade` as the in-place route and the 0.11.x release for v8 and v9.

## Tests and docs

- `migrations.rs`: the served range and birth stamps. `upgrade/tests.rs`: the synthetic v6 and legacy v8 routes end at v10 with spellings and branches kept, a new v9 → v10 case, the retired-ancestry case runs the route and checks the ancestor keeps v8, the already-current case's target table, and the v7 → v8 interruption, history and pending-intent cases continue to v10 before the graph is served. `manifest/tests.rs`: a sub-floor stamp refuses before any effect on both vintages; the genesis probe separates a foreign initializer by lineage since both vintages share one stamp.
- `tests/system_column_upgrade.rs`, `legacy_columns.rs`, `lifecycle.rs`: stamps read 10.
- `crossversion_upgrade.rs`: the fences are `current_v10_*`, the rebuilt-graph check expects 10, and the genuine v0.9/v0.10 journeys run the default route to v10 on their branched fixtures after the `--to-format 8` stop. `ci.yml`: the same names; the `V5 ↔ V9 Format Fence` job is now `V5 ↔ V10 Format Fence` (if that context is in branch protection, it needs the new name).
- `openapi.json`: the health field's description (the binary serves one stamp).
- Docs: `versioning.md` (contract, v10 entry, route text, support matrix), `upgrade.md`, `testing.md`, `ci.md`, `lance.md`, and the v0.12.0 release note under compatibility changes.

Run locally: the canonical workspace test graph (every engine, server, cluster, CLI, compiler, policy and storage suite; the CLI system suites need the server binary at `target/debug` as CI has it; the benchmark provenance unit tests fail here only because they run `git commit` under a signing agent this shell cannot reach), the GQT lib, runner dispatch and corpus, both Clippy graphs, `cargo fmt`, `check-docs.py`, `typos`, `check-agents-md.sh`, `check-workflow-action-pins.py` and `check-storage-upgrade-ci.py --self-test`. The predecessor-binary fences skip locally as documented.

## Next

Step 3 converts mutation and load to detached staging with held-handle promotion and ports the failure-window matrix onto seams; later steps convert ensure-indices, merge and Optimize, each deleting its sidecar kind.


ragnorc:

PR #731, opened 2026-09-16
engine: mutation and load stage detached commits and publish pins (RFC 0067 step 4a)

ragnorc:
RFC 0067 rollout step 4a: mutation and load stage detached commits. Stacked on #728 (the v10 stamp), which is stacked on #727 and #725.

## What changes

- **Every mutation and load table effect is a detached commit of the pinned base.** `commit_all` no longer arms, confirms or deletes a recovery sidecar and never moves a table's linear HEAD: it commits each staged transaction detached, publishes every pin as `(base + 1, staged version, transaction uuid)` in the one manifest CAS, and then promotes each pin from the handles it holds (`promote_held`, replaying at `base` through the sealed `promote_detached`), best effort. A failure before publication leaves the graph unchanged and the detached manifests as reclaimable garbage; a failure after publication leaves a pending pin that reads serve through the staged version and the next writer of the table, or `cleanup`, promotes. The strict-insert conflict re-probe, the effect-free sidecar retirement and the exact-commit outcome are gone with the sidecar.
- **Writers open the pin, never HEAD.** `open_pinned_for_write` serves the held handle for the pin, otherwise resolves it through the read-handle cache and promotes a pending predecessor first, so a write stages from a linear base whenever promotion is healthy. A pin whose promotion is blocked (a foreign linear commit occupies its target) stages the write from the detached version and the write's own promotion waits behind the block, as RFC 0067 §Promotion specifies; a linear pin whose HEAD moved without a publication still refuses through the existing baseline check.
- **The writers that still commit on the linear HEAD promote a pending pin first.** `promote_pin_at_head` judges the pin from the HEAD such a writer opens anyway: a HEAD at the pin's version is checked by its recorded transaction with no request, a HEAD behind it promotes and reopens, a HEAD beyond it asks the target once. Wired into branch merge (existing target, pure-insert source recheck), index maintenance, schema apply, the system-column upgrade, Optimize and repair; the two first-touch fork sources (merge, index maintenance) promote the inherited pin through `promote_inherited_pin`, free when this process holds the twin. `merge_cost.rs` keeps its delta-scoped open count. A blocked pin refuses those writers with a manifest conflict naming `omnigraph repair`, because their linear effect would rebase over the foreign commit.
- **`omnigraph repair` reports blocked pins.** A new `blocked_promotion` classification, always `refused` and never adopted with `--force`; a pending pin is promoted before drift is judged, so a promoted twin reads as `no_drift` instead of being classified as the foreign commit it is not.
- **Seams.** `mutation.post_arm_pre_effect`, `mutation.post_sidecar_pre_fork`, `mutation.sidecar_confirm_put` and `mutation.sidecar_post_publish_delete` are gone; `mutation.post_table_commit` and `mutation.post_fork_pre_commit` keep their names over the detached effect; `mutation.post_publish_pre_promotion`, `promotion.post_landed` and `cleanup.pre_reap` name the new windows.
- **Retired with the sidecar.** `TableStorage::reopen_for_mutation`, the strict source ids on a staged write, `delete_sidecar_after_publish`, `is_retryable_commit_conflict`, the `guarded` seam helper, and `StagedTableEntry::pending_mode`. `new_occ_sidecar_v9` and `confirm_occ_sidecar_v9` are `cfg(test)`, kept for the classifier's own tests until the classifier is removed.

## Tests

- New owner `tests/detached_commit_matrix.rs`: writer (insert, multi-table, cleanup) × window (after each detached effect, before and after publication, between and inside promotions, before cleanup's reap) × fault (error return, process kill, race with a concurrent insert) × recovery actor (same handle, fresh handle, other process, cleanup, nobody), one oracle. The default run covers every writer, window and fault with the fresh-handle and read-only actors; `OMNIGRAPH_MATRIX=full` adds the rest.
- `failpoints.rs`: `linear_writers_promote_a_pending_pin_first` (index maintenance, merge with pending source and target pins, Optimize, repair, schema apply, each over a pin whose promotion was interrupted) and `blocked_promotion_keeps_writing_detached_and_repair_reports_it` (a foreign commit at the target: the next mutation chains detached, index maintenance refuses, forced repair reports and adopts nothing, a fresh handle reads both writes); the mutation and load sidecar recovery tests (Armed intents, effect-free retirement, roll-forward of mutation and load residue, the issue 602 stale-sidecar heals) are retired with the sidecar; `interrupted_write_leaves_main_writable_without_recovery_issue_554`, `first_touch_fork_failure_before_effects_leaves_no_residue` and `inline_delete_conflict_rejects_without_effect` pin the detached-path behaviour; the recovery-machinery tests (discovery races, the cleanup barrier, the storage-fault contracts, in-process healing, the follower barrier) take their pending sidecar from the index writer.
- S3 (RustFS in CI): `s3_write_pending_pin_is_promoted_by_the_next_write` (a load interrupted after publication, promoted by the next load) and `s3_recovery_sidecar_lifecycle_heals_on_the_next_write` (the index writer's sidecar put, listed, rolled forward and deleted through the S3 adapter) replace the retired mutation-sidecar S3 cell.
- `write_cost.rs`, `warm_read_cost.rs`: a write writes and deletes no control object and opens nothing on a warm handle; a read after a write is served from the version the writer landed.
- GQT: the four cases arming the retired seams are gone; `mutation_pending_pin_survives_reopen.gqt` (two seams on one mutation, a pending pin read after reopen and promoted by the next write) replaces the two-seam owner; `dst_mutation_failure_keeps_writing.gqt` loses its known failure; `merge_armed_intent_blocks_main_until_reopen.gqt` (a merge that fails after its effects and again inside its error recovery leaves the documented rollback-eligible intent that refuses the next write on main until a read-write reopen) is the corpus's known-failure case and the runner's fixture for it.
- Docs: `writes.md` (the detached variant of the protocol, blocked pins), `recovery.md`, `testing.md`, `operations/maintenance.md` (`blocked_promotion`), release note.

- `crates/omnigraph-dst`: the crash-window catalog follows the retired mutation seam; the cost golden is regenerated (every mutation-class op loses its sidecar adapter PUTs and DELETEs and the ref LISTs its arm paid, and pays one Lance manifest PUT per touched table for the twin; Cleanup reaps the detached manifests); four pins whose shape needs a mutation sidecar retire (a stale confirmed sidecar blocking maintenance, a lost sidecar confirm write, a corrupt sidecar consumed on first contact, raw-row arbitration of failed attempts); the issue-554 panel keeps its availability guarantee and no longer requires a deferred-recovery row; the bounded-staleness and ack-loss pins take new seeds from two new search instruments, and the ack-loss pin now expects exactly-once application (a retried keyed insert meets its own row).

## Notes

- Two pre-existing engine/model disagreements surfaced by the shifted DST schedules, out of scope here: under the bounded-staleness realm a stale absence of `_schema.ir.json` or `__schema_state.json` makes an op refuse with "graph schema contract is incomplete; manual coordination is required", and a zombie `__recovery/` listing makes a write refuse with `RecoveryRequired` for a sidecar that no longer exists. Roughly a third of the searched seeds hit one of them.

- `omnigraph-cluster --test failpoints::schema_apply_error_before_graph_movement_removes_sidecar` overflows the 2 MiB test-thread stack in a local debug build before and after this change; CI runs every test job with `RUST_MIN_STACK=16777216`, where it passes.


ragnorc:

PR #734, opened 2026-09-16
engine: branch merge stages detached chunk chains and publishes pins (RFC 0067 step 4)

ragnorc:
RFC 0067 rollout step 4, third writer: branch merge stages detached commits. Stacked on #733 (ensure-indices), which is stacked on #731, #728, #727 and #725.

## What changes

- **Every merged table's chunks chain as detached commits of the pin captured before classification.** `prepare_existing_merge_target` opens the pin (`open_pinned_for_write`, promoting a pending predecessor first, staging behind a blocked one); each chunk of a proven insertion chain, an adopted delta or a three-way rewrite commits through `commit_staged_detached` on the previous chunk (`MergeChain`, one link per chunk within the existing `MAX_BRANCH_MERGE_DATA_TRANSACTIONS` ceiling, now checked by `plan_merge_chain_length`); the table's pin publishes `expected + chain length` with the tip as its staged version, in the one manifest commit. The writer then promotes every link in order (`HeldPromotion` gained a `chain`, and `promote_held` replays the links from the base, continuing from a twin a racing promoter landed).
- **No sidecar, no pre-minted identities, no in-process compensation.** `new_branch_merge_sidecar_v9` is `cfg(test)`, `confirm_branch_merge_sidecar_v9` and `recover_failed_branch_merge_under_gates` are deleted, the exact-commit outcome check and the `RecoveryRequired` wrapping are gone; a merge that fails before publication returns its error and leaves only detached chains and unreferenced forks, which cleanup classifies. A first-touch fork on a named target is created under the gates with no intent record. Pointer adoption copies the source's entry as it is, so a pending pin becomes the target's pin too.
- **Seams.** `branch_merge.post_sidecar_pre_fork`, `branch_merge.post_effects_pre_confirm` and `branch_merge.pre_error_recovery` are gone; `branch_merge.post_fork_pre_commit`, `branch_merge.post_table_effect` and `branch_merge.post_publish_pre_promotion` are new; the chunk seams and `post_authority_capture`, `post_candidate_validation`, `post_phase_b_pre_manifest_commit` keep their names.
- **A first-touch fork accepts an inherited handle opened at a pending pin.** The mutation path's pre-fork check compared the inherited handle's version with the manifest's published version; a handle opened at a pending pin resolves to the staged version, which the pointer-adoption scenario below exposed. The check now accepts the pin's staged version; the fork loop already promoted the pin before forking.
- **The classifier keeps the `BranchMerge` kind** for sidecars written before the change, as #731 and #733 did for theirs; step 5 removes them together.

## Tests

- `detached_commit_matrix.rs` gains `Writer::Merge` (a branch holding the row under test, merged into main; the same windows, faults and recovery actors; a race with a main insert makes the merge lose its CAS).
- `failpoints.rs`: twenty-five sidecar-era merge tests retire (phase-B recovery on reopen, pointer-slot replay, sidecar delete failure, rollback and restore, armed-prefix rollbacks, first-touch ref recovery, net-zero confirmation, the v1-sidecar classifier synthesis). New: every chunk window and the table and pre-publish windows leave no residue and the retry lands (`assert_partial_merge_leaves_no_residue` over the adopt and rewrite paths, plus the two multi-chunk fixtures); an interrupted promotion leaves a pending chain of two that the next write promotes link by link; a target advance after the detached effects loses the merge's CAS; a foreign linear commit after the effects blocks promotion, not the merge, and later writes chain behind it; a first-touch fork failure leaves garbage and the retry owns its own fork; pointer adoption carries a pending pin. The pre-arm fences (ABA, sync, source advance, late uncovered drift, pointer adoption) are unchanged. The two issue-694 live-compensation tests and the merge-sidecar body re-read test retire with the compensation path; the follower-attribution test only requires its Optimize producer to advance the table. The foreign-commit test documents that a conflict-free replay over a foreign commit may land as unreferenced history one past it while the pin still names the chain's tip.
- The recovery-machinery tests (discovery races, the cleanup barrier, the storage-fault contracts, in-process healing, the S3 sidecar cell) take their pending sidecar from Optimize (`graph_with_optimize_work` + `leave_confirmed_optimize_sidecar`), the last writer that arms one that survives the next step.
- `write_cost.rs`: a merge writes and deletes no control object; `merge_cost.rs` keeps its delta-scoped open count.
- GQT: the issue-694 cases move to `branch_merge.post_fork_pre_commit` and `branch_merge.post_table_effect`; `merge_armed_intent_blocks_main_until_reopen.gqt` and the two runner tests that needed a corpus case with a genuine `RecoveryRequired` retire, because no corpus writer arms a recovery intent any more (the marker stays parsed and refused until the classifier retires); the runner fixtures rename the seam.
- `crates/omnigraph-dst`: the crash-window catalog follows the renamed and new merge seams (the mutation and index renames land on #731 and #733).
- Docs: `writes.md` (writer row and the merge's protocol), `recovery.md`, `testing.md`, the RFC 0045 example, release note.

## Review regression tests (17 September 2026)

Seven `failpoints.rs` cases pin the boundaries the stack review reproduced, five of which failed before the fixes now in #727: the single-link pending-pin diff control, the blocked-chain and pending-merge diffs, cleanup on a blocked pin, the aged detached-manifest reaper, the reader racing a promotion, and a stale handle writing after another handle's promotion (the cause of the intermittent 409s in the server's concurrent-insert test, green eight of eight afterwards). The merge phase topology fix for the bench crate and the maintenance guide's blocked-pin sentence (branch merge and index maintenance stage behind a block) are here too.


ragnorc:

PR #735, opened 2026-09-16
engine: schema apply stages detached rewrites and publishes its contract without a sidecar (RFC 0067 step 4)

ragnorc:
Stacked on #734 (branch merge). RFC 0067 rollout step 4, fourth writer: schema apply.

## What changes

- **Rewrites stage detached.** An existing-table rewrite is a detached Overwrite of the promoted HEAD, published as a pin `(published + 1, staged version, transaction uuid)` and promoted after the manifest commit, exactly like the other writers. Schema apply still promotes every pending pin of every live table before it plans and refuses a blocked one, since it is graph-global.
- **Added types reclaim, not refuse.** A new type is still a linear version-one create at its identity path. That path is a deterministic function of the accepted identity allocator (`next_identity_id`), so the RFC's "fresh incarnation path" wording was wrong: a crashed attempt leaves its dataset exactly where the retry creates it. Nothing references an unregistered incarnation path, so the retry deletes the leftover under the schema sentinel and creates. The RFC row is corrected.
- **The contract's authority is the publishing commit.** `__schema_state.json.staging` gains `publication: { graph_commit_id, parent_commit_id }`. Recovery promotes the staging when that commit is in main's lineage and discards it otherwise; property-only migrations are no longer ambiguous and the table-set discriminator is gone. The writer installs the live contract from memory after the commit, so an open in another process discarding its staging cannot tear the graph. A read-only open refuses a published-but-uninstalled contract (the cluster test's wording still matches) and serves an unpublished staging as absent.
- **No sidecar.** A failure before the manifest commit returns the plain error; only a failure after it reports `RecoveryRequired`, naming the published commit. The same handle finishes the installation on its next write through a flag set by the failed apply (no per-write staging probe: the mutation cost contract still pins four existence probes); other handles and processes converge at their next read-write open, which also reclaims a sentinel left by a crashed apply. Before this change an apply that died after deleting its sidecar left the sentinel forever.
- **System-column upgrade unchanged.** RFC 0040's writer keeps the v9 exact protocol, unmarked staging and sidecar; `new_schema_apply_sidecar_v9` is test-only.

## Lance fence

`rfc_0067_replay_over_its_own_twin_is_refused_only_for_self_conflicting_kinds` gains the Overwrite case: Lance accepts the twin and returns the first commit's version without adding one, so racing promoters of a schema-apply pin leave no residue.

## Tests

- `failpoints.rs`: pre-commit crash discards staging on reopen; partial promotion after a commit crash (read-only refusal, mixed rename state completed); retry over its own unpublished staging in-process; load after a pre-publish failure keeps the accepted catalog; pre-staging failure leaves no residue (no rollback publication, HEAD at the pin); metadata-only before/after staging; post-publish failure healed by the next write on the same handle; retry reclaims an abandoned add-type dataset; rename+rewrite and two-table partial effects leave no residue; losing the manifest CAS to a concurrent publication leaves no residue and the retry plans from the winner. Retired: sidecar rollback-retry, promotion-prepass, foreign first-touch winner, phase-B roll-forward, same-table fail-closed and the Optimize late-schema-sidecar recheck, whose producers no longer exist.
- `detached_commit_matrix.rs`: `Writer::SchemaApply` (nullable property on Person); the sentinel refuses the race insert. Full run: 52 cells green.
- `write_cost.rs`: `schema_apply_writes_no_control_object` (six contract writes, three staging deletes, no sidecar).
- `forbidden_apis.rs` registry rows updated; DST catalog lists 72 windows.

## Cost

The mutation path is unchanged (`write_schema_io_is_bounded_to_capture_fence_and_effect_gate` still pins four existence probes; the same-handle heal runs only behind the flag). A read-write open lists the branch refs once more to reclaim a stale sentinel, and a read-only open probes `__schema_state.json.staging` once for the coherence proof; the DST cost golden records exactly those deltas (`_audit` exists +2, `_verify` exists +13, one list per open).

## Validation

Local: engine suite with failpoints, workspace rest, GQT, DST suites, `lance_surface_guards`, fmt, both clippy graphs, docs/typos/agents/pins checks.


ragnorc:

PR #740, opened 2026-09-17
engine: the system-column upgrade stages detached renames and arms no sidecar (RFC 0067 step 5)

ragnorc:
Stacked on #739 (Optimize). RFC 0067 rollout step 5, first half: the last sidecar writer moves onto detached commits, which is what lets the next PR delete the recovery machinery.

## What changes

- **The RFC 0040 system-column upgrade takes schema apply's shape.** Each table's rename-only `Project` commits detached from its promoted pin and is published as a pin `(published + 1, staged version, transaction uuid)`; the staged schema contract names the publishing graph commit; one manifest CAS under the exact graph head publishes every pin; the contract installs from memory; the held renames promote. The upgrade still takes the schema gate, the schema-apply sentinel, main's branch gate and every table gate, promotes every pending pin before it plans and refuses a blocked one.
- **No sidecar, no stamp advance.** Since v10 both column vintages share one `__manifest` stamp, so the old protocol's stamp advance was already a no-op (`from == to == 10`); the vintage is the contract's `system-columns` feature. The report keeps `stamp_before`/`stamp_after` for the CLI contract.
- **Failure contract.** Before the manifest commit: the plain error, the graph unchanged, the sentinel released; the next read-write open discards the unpublished staging (and reclaims a dead writer's sentinel); a read-only open serves the legacy contract; rerun the upgrade. After it: `RecoveryRequired` naming the published commit; the next read-write open, or the same handle's next write entry, installs the contract; a read-only open refuses until then.

## Lance fence

`rfc_0067_replay_over_its_own_twin_is_refused_only_for_self_conflicting_kinds` gains the rename-only `Project` case: Lance's `check_project_txn` refuses `Project` against `Project`, so a racing promoter's replay is refused and adds no version.

## Tests

- `system_column_upgrade.rs`: four pre-publication windows leave no residue (plain error, no sidecar, sentinel released, HEAD and pin unmoved, read-only open serves the legacy graph, retry completes); post-commit failure finished by the next open (read-only refusal in between); same-handle heal; skipped promotion leaves pending pins the next writer and cleanup promote; same-handle retry after a pre-effect failure; a dead writer's sentinel reclaimed at open; the control-object cost (six contract writes, three staging deletes, no sidecar). Retired with their producer: the interrupted-recovery and lock-reclaim-crash cells of the roll-forward executor.
- `forbidden_apis.rs` rows for the upgrade module; seam `schema_apply.post_sidecar_pre_effect` renamed `schema_apply.post_lock_pre_effect`, `system_column_upgrade.after_stamp_advance` removed.

## Validation

Local: the workspace test graph with failpoints, the engine library tests, failpoints, the failure matrix, recovery, maintenance, legacy columns, schema apply, the upgrade owner, registry and seam-name guards, `lance_surface_guards`, DST from its crate directory, fmt, both clippy graphs, docs/typos checks.


ragnorc:

PR #743, opened 2026-09-17
dst: remove the keep-serving watch machinery (RFC 0067 follow-up)

ragnorc:
Stacked on #741 (recovery machinery removed). Harness-only follow-up: the DST keep-serving watch could only be entered by a live handle refused on a pending recovery sidecar, and no writer arms one any more.

## What is removed

- `Scenario::keep_serving_ops`, the watch (`KeepServingWatch`), the interrupting-op widening (`WatchInterrupt`, `CompositionHypothesis`, `composition_hypotheses`, `op_modes`, `WatchRuling`, `reconcile_watch_resolution`, `resolve_keep_serving_watch`), the `keep-serving-*` report row prefixes, and the six arming/resolution sites woven through the universe loop. About 900 harness lines.
- The `LiveWriteAvailability` oracle and its detector (census: 21 oracles, 24 detectors, regenerated).
- The clean-recovery-state legality carve-out (`is_recovery_barrier_rejection`) and its `reopen-heals-barrier@` row: the engine no longer emits "requires a clean recovery state" anywhere.

## What stays

- `reconcile_after_failure` and its single-op hypothesis arbitration are untouched: every failure still reconciles in its own iteration, which is exactly why a second unjudged op can no longer exist.
- The issue-554 panel survives as `dst_availability_panel_survives_harsh_faults_issue_554`: the same 13 seeds under the same harsh fault plan (`error_pct` 80, Lance realm), and every seed must run green — any red is a genuine availability or oracle regression, no longer a designed wedge.

## Evidence

The cost golden is byte-identical, which is the knob's standing constraint ("must never perturb an existing pin's op stream or rng draws") holding through its removal. Full DST suite green from the crate directory (46 scenarios), DST clippy clean, fmt clean. Engine crates untouched.


ragnorc:

